Issue #17, opened 2025-04-30
Groups | UI Foundations


bdebut01:
**In scope**:
- Pages matching CRUD operations: Groups details, edit, delete

**Out of scope**
- Permissioning based logic: Group moderator vs. regular member view difference

PR #20, opened 2025-05-05
Core Item visibility permissions

rjendoubi:
This PR gets us the core authorization mechanics of restricting which Items are visible to which users, but no integration into the Views or UI yet.

The separate permission of _borrowing_ Items (as opposed to just _viewing_ them) is likewise not implemented here! Concentrating first on which Items users should be able to "see" / access in the first place. That said, 1) it'll be a relatively straightforward extension to this base work and 2) as far as I can imagine right now, there should be no reason why a user can "see" an Item but not be entitled to borrow it, assuming it's available.

*    Additional basic admin attributes for Group Memberships.
    
    `joined_at`, `status`, `status_changed_at`, `status_changed_reason`.
    For suspending / banning from groups.

*    Start of tests around Group membership.
    
    Using the standard Django `unittest` style tests for now. I prefer
    `pytest`, but there's something to be said for sticking with the
    defaults, especially under time pressure.

*   BorrowdGroup.remove_user, plus tests.

*    Add attribute Item.trust_level_required, plus custom permissions.
    
    Custom `*_this_*` permissions will be used later to facilitate
    object-level auth.

*    Bring in django-guardian and integrate with our user and group.
    
    `django-guardian` is the most flexible and comprehensive solution today
    for object-/record-level permissions in Django.
    
    Django's OOTB auth system only allows for type-/class-level permissions:
    i.e. "X group is allowed to perfom Y operation on Z object-type." What
    it doesn't let you say is, "X group is only allowed to perform Y
    operation on THIS INSTANCE of object-type Z." This is critical
    functionality for us, where object access is dependent on group
    membership. `django-guardian` builds on extension points built into the
    Django framework, providing helper functions for easily assigning and
    checking arbitrary permissions on a per-object-instance basis.
    
    Another option considered was to simply query the
    user->group->user->item relationships directly. While simpler, and
    easier to keep updated, there is no obvious way to accommodate
    different permission types with this approach. One would quickly need to
    implement custom "Role" or "Permission" classes - eventually building
    something that looks very like the core Django auth model and/or
    django-guardian itself. Therefore, at the cost of a bit more
    complexity, django-guardian provides a readymade, generalized approach
    to this problem area which feels like it has a longer runway than
    building (not to say "hacking") our way there organically over time.

*    Leverage django-guardian with signals for permission management.
    
    Doing the permission management work in Django signals makes it
    significantly easier to comprehensively cover all the bases: we don't
    have to ensure that objects are only made via a certain method or in
    a certain way, we simply catch all creates/updates and update
    permissions right alongside.


borrowd_groups/tests/test_remove_users_from_groups.py
```diff
@@ -0,0 +1,80 @@
...
+
+        ## Create a group without adding the user
+        group: BorrowdGroup = BorrowdGroup.objects.create(
+            name="Group", created_by=user, updated_by=user
+        )  # type: ignore
+
+        # Act
+        ## Attempt to remove the user from the group
+        # TODO: Should this raise an exception?
+        group.remove_user(user)
```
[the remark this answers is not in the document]
ajk8: I could go either way on this (raise an exception or log an error). The two reasons why I might not raise an exception are:

1. I wouldn't want to use it as control flow
2. I wouldn't want to get spammed with alerts on it -- the operation itself isn't a problem per se, because the outcome is the same whether the user was a member of the group or not beforehand

That said, I expect that it would be a pretty uncommon case, so if you'd like to have it raise an exception, then 👍

borrowd_items/apps.py
```diff
@@ -4,3 +4,10 @@
 class BorrowdItemsConfig(AppConfig):
     default_auto_field = "django.db.models.BigAutoField"
     name = "borrowd_items"
+
+    def ready(self) -> None:
+        # Unfortunately this unusued import is the recommended
+        # approach when using the `@receiver` decorator; see
+        # section "Where should this code live?" in the docs:
+        # https://docs.djangoproject.com/en/5.1/topics/signals/
+        import borrowd_items.signals  # noqa
```
ajk8, line 13: I agree, this pattern is 🤢. Imports during initialization get pretty gnarly in django!

borrowd_items/models.py
```diff
@@ -26,9 +34,41 @@ class Item(Model):
...
         return self.name
 
     def get_absolute_url(self) -> str:
         return reverse("item-detail", args=[self.pk])
+
+    class Meta:
+        # Permissions using the naming conventon `*_this_*` are used
+        # for object-/record-level permissions: whereas the permission
+        # `view_item` would allow a user to view "any" Item, the
+        # permission `view_this_item` allows viewing a specific Item.
```
ajk8, line 56: Super helpful comment!

borrowd_items/signals.py
```diff
@@ -0,0 +1,33 @@
...
+        # Assign view permissions to all Groups of which the owner
+        # is a member and has an equal or greater Trust Level than
+        # the level required by this Item.
+        # mypy struggling a bit here again, even though intellisense
+        # knows that Item.owner (BorrowedUser) does indeed have a
+        # a Groups accessor.
+        allowed_groups = instance.owner.groups.filter(  # type: ignore[attr-defined]
+            membership__trust_level__gte=instance.trust_level_required
+        )
+        assign_perm("view_this_item", allowed_groups, instance)
```
ajk8, line 33: When do we assign `borrow_this_item`?

ajk8 (approved):
A few minor things, but I would be fine moving forward with this.

borrowd_groups/tests/test_remove_users_from_groups.py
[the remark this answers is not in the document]
rjendoubi: Will track in #22 :+1:

borrowd_items/signals.py
rjendoubi: Not considering that yet - have expanded PR description to clarify that - but in all likelihood, yes, that perm would also get assigned in all the same situations :+1:

rjendoubi:
@bdebut01 Made the requested change, and rebased following #17's merge to `main` :+1:

bdebut01 (approved):

PR #23, opened 2025-05-05
Bugfix: Group template lookup

rjendoubi:
This work comes after #20, please let's review and merge that one first!

Small, pedantic changes, fixing some hassle I created earlier :pray: 

*    Fix signup button on landing page for new button component.

*    Fix lookup of BorrowdGroup template names.
    
    It's verbose to have to type "borrowdgroup_foo" for all group template
    names. The only reason our Group model is called `BorrowdGroup` is to
    avoid clashing with the Django core group model. We don't need to spell
    it out everywhere.
    
    This fixes a bug I introduced in a82b24e where, with the new
    `BorrowdGroup` class name, the template lookup mixin no longer worked
    correctly.


bdebut01 (approved):
Whoops, conflict resolution fail on my part wrt that button! Thanks for fixin that. Rest looks good!

rjendoubi:
Thanks @bdebut01 - would you mind re-approving? Apparently this is required because the other PR got merged in the meantime?
>  Review from bdebut01 is stale because it was submitted before the merge base changed.

ajk8:
> Apparently this is required because the other PR got merged in the meantime?

Does this have all the changes from the other branch as well? It's really hard to tell what's distinct about this PR.

bdebut01 (approved):

rjendoubi:
@ajk8 Yes, it had the changes from the other PR as well. I thought that once that PR was merged, this would show only the latest changes, but I see that's not the case - so yes, pretty impossible to see what this adds :scream:

I'm going to see if there's a way to update it; otherwise will close / delete this PR and re-raise (merging is still blocked, despite Ben's latest approval, which doesn't make sense - it may have gotten into some bad state??~)

rjendoubi:
So apparently what I needed to do was rebase on latest `main` :+1: 

@ajk8 hopefully the changes make more sense now (i.e. are discernible at all :laughing:). Pls approve if it looks good?

ajk8 (approved):

PR #26, opened 2025-05-06
Foundations for Items CRUD UI: detail, create/edit, delete, list

bdebut01:
+ Tweak UI on Profile page to more clearly delineate items from rest of UI

### DIFF
--- a/borrowd_items/views.py
+++ b/borrowd_items/views.py
@@ -9,6 +9,7 @@ from django.views.generic import (
     UpdateView,
 )
 
+from borrowd.models import TrustLevel
 from borrowd.util import BorrowdTemplateFinderMixin
 
 from .models import Item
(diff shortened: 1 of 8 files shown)


borrowd_items/views.py
```diff
@@ -20,6 +21,8 @@ class ItemCreateView(BorrowdTemplateFinderMixin, CreateView[Item, ModelForm[Item
 
     def form_valid(self, form: ModelForm[Item]) -> HttpResponse:
         form.instance.owner = self.request.user  # type: ignore[assignment]
+        # default trust level for now
+        form.instance.trust_level_required = TrustLevel.LOW
```
ajk8, line 25: Does this silently set the trust level to low? If so, are we sure that's the behavior we want? Can address in a later PR, but wanted to raise the question.

ajk8 (approved):
One question that can be deferred. Looks good otherwise.

borrowd_items/views.py
bdebut01: Yeah, was mainly setting that to let the processing go through (it's a required field). It'll be replaced with the real logic the next time we do anything to `items` so very much a temporary line there.

bdebut01:
Sorta annoying to require round 2 approval on direct PR feedback commits... but @ajk8 @rjendoubi seeking that to merge!

ajk8 (approved):

Issue #27, opened 2025-05-06
Group Invites: QR codes


rjendoubi:
https://pypi.org/project/django-qr-code/

ik-brwd:
A terrible mockup I take full responsibility for: 

<img width="1062" height="751" alt="Image" src="https://github.com/user-attachments/assets/7055f5a2-3622-46fa-9b48-37b6cc014e12" />

PR #28, opened 2025-05-07
Group Invite flow

rjendoubi:
Implements the mechanism for joining Groups, via an Invite link.

Not yet implemented:
* Securing initial access to the Groups themselves
* Removing users from Groups

The process goes:
1. Invitor (current Group member) clicks "Get Invite Link"
2. Invite link is displayed (currently on separate page; would be easy to make this a modal). Select or use button to copy to clipboard.
3. Invitor gives invite link to Invitee
4. Invitee opens invite link (GET), which displays Join Group page
5. Invitee selects a Trust Level and confirms Join Group (POST)
6. Invitee becomes a Member of the Group

The URL for _generating_ the invite link is consistent:

  /groups/n/invite

The URL for _joining_ a group, i.e. the invite link that was generated, consists of an opaque token:

  /groups/join/<token>

Signing / encrypting / securely obfuscating invite links is in line with Borrowd's core value of Trust: since Groups should be hidden without explicit invitation, this approach prevents enumeration attacks on Group names and IDs.

Currently the only information encoded in the join link token is the Group name (user-defined) and ID (autoincrementing database pk), which must match each other. We would have the option of later encoding other information, e.g. the user who sent the invite, or making invite links specific to certain invitees (by checking against recipient email / username.

Note also invite links are currently time limited to 1 week. We can make this shorter, longer, or remove altogether.


bdebut01:
At the risk of sounding like a non-technical _shmo_... if we wanted "tiny"/shortened urls that'd require database persistence yeah? 

Generate slug mapped to full url above (so we could encode the various attrs you mention/leave open for placing more), store in db, lookup on user visit?

Just want to raise the topic for discussion at least. I'm a man that likes shorter URLs lol, not that anyone is manually typing urls at this point.

borrowd_groups/views.py
```diff
@@ -42,18 +84,131 @@ class GroupDetailView(BorrowdTemplateFinderMixin, DetailView[BorrowdGroup]):
...
+    descriptive error page if neccessary.
+
+    On GET, displays basic information about the Group and a button
+    to confirm joining.
+
+    Then on POST, actions the joining of the user into the Group and
+    displays a confirmation.
+    """
+
+    def _validate_invite(
```
bdebut01, line 116: Not sure if by "Not yet implemented: Securing initial access to the Groups themselves" encompasses this, but figure we may as well do a check in here for Group **not** "requiring membership approval" alongside the rest of the validation. And future development can pick up the other half/workflow that covers Groups that do require approval.

templates/groups/group_join.html
```diff
@@ -0,0 +1,17 @@
+{% extends "groups/layout.html" %}
+{% block head_title %}Join '{{ object.name }}'!{% endblock %}
+
+{% block groups_content %}
+<c-box>
+     <div class="flex flex-col h-full">
+        <c-h2>Join Group?</c-h2>
+        <c-h1>{{ object.name }}</c-h1>
+        <p class="text-gray-700 mb-4"><b>Description:</b> {{ object.description }}</p>
+        <p class="text-gray-700 mb-4"><b>Membership requires approval:</b> {{ object.membership_requires_approval|yesno:"Yes,No" }}</p>
```
bdebut01, line 10: Related to other comment about Groups Requiring Approval, but for now maybe we can just switch the `Join` text to `Request to join` if the group does require it. And send that request to dev/null for now. Groups that _don't_ require approval shouldn't need to display anything here.

templates/groups/borrowdgroup_detail.html
```diff
@@ -12,10 +12,14 @@
         <c-h1>{{ object.name }}</c-h1>
         <p class="text-gray-700 mb-4"><b>Description:</b> {{ object.description }}</p>
         <p class="text-gray-700 mb-4"><b>Membership requires approval:</b> {{ object.membership_requires_approval|yesno:"Yes,No" }}</p>
-        <c-button @click.prevent="onShareClick">Share Invite Link</c-button>
+        <c-button>
+            <a href="{% url 'borrowd_groups:group-invite' object.pk %}">
```
bdebut01, line 16: On second thought, why don't we just offer **only** a "Copy Invite Link" button and skip the new page/modal and everything else. They're already looking at the Group details so path of least resistance is just a direct copy. Can't think of the use case for actually showing them a link... unless they're standing right next to who they want to share with but even then the url is so long you wouldn't copy by looking at it. (You'd maybe offer a QR code tho 😝 )

borrowd_groups/views.py
```diff
@@ -42,18 +84,131 @@ class GroupDetailView(BorrowdTemplateFinderMixin, DetailView[BorrowdGroup]):
...
+
+        # Just rename the var to avoid confusion
+        group: BorrowdGroup = val_res
+
+        form = GroupJoinForm(request.POST)
+        # Making sure a Trust Level has been selected
+        if not form.is_valid():
+            return render(request, "groups/group_join.html", {"form": form})
+
+        # Wonder what we can do to cast request.user to our
```
bdebut01, line 190: Right... at some point we're gonna have so many `type: ignore[...` the `mypy` check becomes useless? Just don't want to create too much tech debt for ourselves so early. But then again, what's early stages of any project creation if not mass-tech-debt printing 😅

bdebut01 (changes requested):
Some comments for discussion! If any look like they're expanding scope too much though, fine to proceed with just the minor code changes. And punt to a Task on the Board

borrowd/config/base.py
```diff
@@ -162,6 +162,11 @@
 
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
 
+#
+# Borrowd settings
+#
+BORROWD_GROUP_INVITE_EXPIRY = 60 * 60 * 24 * 7  # 1 week
```
ajk8, line 168: Love the clear math instead of a big integer that leaves the reader to figure it out. One nit here: could this be renamed to `BORROWD_GROUP_INVITE_EXPIRY_SECONDS`? That just makes it all the more clear when using the variable.

templates/groups/borrowdgroup_detail.html
```diff
@@ -12,10 +12,14 @@
         <c-h1>{{ object.name }}</c-h1>
         <p class="text-gray-700 mb-4"><b>Description:</b> {{ object.description }}</p>
         <p class="text-gray-700 mb-4"><b>Membership requires approval:</b> {{ object.membership_requires_approval|yesno:"Yes,No" }}</p>
-        <c-button @click.prevent="onShareClick">Share Invite Link</c-button>
```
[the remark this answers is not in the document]
rjendoubi: Good catch, done :heavy_check_mark:

borrowd_groups/views.py
rjendoubi: Ah - divided by a common language - you're correct in that the functionality you describe will have to come in a later PR, but that's not even what I meant by "displays a confirmation" - I just meant the little "Success" toast :grimacing: 

Have opened #29 for this :+1:

templates/groups/group_join.html
rjendoubi: Mm, I think there will always need to be this "confirm joining" step, because A) we need users to select and submit a trust level, but also B) in general it's safer than performing the join action simply from a `GET` request; it's easier to trick a user's device into loading a random URL than getting them to `POST` to it.

I take the point though that this text should make clear whether approval is required. Will add a comment to the template so this is done later :+1:

templates/groups/borrowdgroup_detail.html
rjendoubi: Honestly yes, displaying a QR code (see #27) is the main reason for this intermediate step. Also because I've experienced a fair few sites where the friendly "Copy to clipboard" button simply and frustratingly _doesn't work_... so I like having the ability to manually c&p as a backup :see_no_evil: #optimizeforanipatterns....

On the happy path though, are you happy to keep it on the basis that we'll have a QR modal? I don't think it should be that hard to add (he said naively)...

templates/groups/group_join.html
bdebut01: Oh, I more meant specifically that "Membership Requires.." line not the full Join page (which agreed we definitely need). For the "Membership Requires..." line I was saying we _didn't need the line at all_ if the Group **doesn't** require approval, and otherwise do that "Request to join" or some other flavor of making it clear they are just requesting to join not actually joining. But can wrap that in to the "Approval flow" task you created

bdebut01 (approved):
Any thoughts on [url shortening]?(https://github.com/borrowd/borrowd/pull/28#issuecomment-2858887741)

ajk8:
> Any thoughts on [url shortening]?([#28 (comment)](https://github.com/borrowd/borrowd/pull/28#issuecomment-2858887741))

This seems like something that we could consider later as an enhancement if the long urls are hard to use.

PR #32, opened 2025-05-08
Group creators automatically become moderator members (backend portion).

### LINKED ISSUE #30: Groups: creator must be initial member + moderator

rjendoubi:
Closes #30.

It would be particularly worthwhile walking through this PR commit-by-commit.

**Now includes UI changes!** :sparkles:

PR content (from the commit messages):

*    Make update_user_membership more comprehensive.

*    Create custom exceptions.
    
    Expect that we'll use these more later.

*    Enforce cannot add existing members to same Group.

*    Groups: allow setting owner trust_level at creation.
    
    * When a User creates a Group, they should automatically be a member
    * Every Group member needs to have a Trust Level with that Group
    * The Trust Level is stored on the Membership model, not the Group
    
    --> Therefore, we need a way to _specify_ a Trust Level at the point
    of Group creation, even though the Group model is not expecting a
    `trust_level` arg (since it has no trust_level field).
    
    The way we do this is to customize the `create()` method of the
    ModelManager for BorrowdGroups, wherein we can do some special handling
    of a `trust_level` parameter. That special handling entails popping off
    that argument from the kwargs, if present, because the underlying
    BorrowdGroup model isn't expecting it, but then make use of Python's
    dynamic malleability by simply adding the trust_level back on to
    the _instance_ of the Group after its created.
    
    In a later change, this "smuggled" attribute will be used by a signal
    handler, which will pass it through to the Membership record created
    for the Group owner at the point of Group creation.

*    Group creators automatically become moderator members.
    
    Add post-save signal handler for Groups which, at the point of creation,
    will ensure that the Group's created_by user is made a member with
    moderator status.
    
    Note that this entails updating a lot of previous tests, which were
    written on the (incorrect) assumption that even Group creators would
    need to be manually added to their own Groups - partly because there
    wasn't a good way to express Group Trust Level at the point of
    creation before.

*    Membership: use more modern unique constraint mechanism.

*    Membership: move Exceptions to dedicated files.
    
    More idiomatic, and avoids potential future circular import issues.

*    Add Trust Level choice on Group Creation UI.


borrowd_groups/models.py
```diff
@@ -163,6 +183,9 @@ class Membership(Model):
         help_text="The User's selected level of Trust for the given Group.",
     )
 
+    class Meta:
+        unique_together = (("user", "group"),)
```
bdebut01, line 228: Looks like Django is [recommending](https://docs.djangoproject.com/en/5.2/ref/models/options/#django.db.models.Options.unique_together) using `UniqueConstraint` instead of this deprecated style?

borrowd_groups/models.py
```diff
@@ -23,6 +23,40 @@
...
+
+        # Manually create the BorrowdGroup object which we'll try to
+        # persist. This is the part which would fail if we passed
+        # unexpected args like "trust_level".
+        group: BorrowdGroup = BorrowdGroup(**kwargs)
+
+        # This instance property is not saved to the database, but
+        # is used in the post_save signal to set the trust level
+        # between the group and the user that created it.
+        setattr(group, "_temp_trust_level", trust_level)
```
bdebut01, line 47: hacky and i kinda love it 😂 😈

borrowd_groups/signals.py
```diff
@@ -2,8 +2,37 @@
...
+    """
+    if not created:
+        return
+
+    group: BorrowdGroup = instance
+    # mypy error: Incompatible types in assignment (expression has type "_ST", variable has type "BorrowdUser")  [assignment]
+    creator: BorrowdUser = group.created_by  # type: ignore[assignment]
+    # By default, assume High trust for a Group which a user has
+    # created themselves.
+    trust_level: TrustLevel = (
```
bdebut01, line 27: See Slack [discussion](https://borrowdworkspace.slack.com/archives/C08JN13HFKP/p1746713942174249)

bdebut01 (changes requested):

borrowd/__init__.py
ajk8, line 1: Nit: could this be moved to `borrowd/exceptions.py`? I believe that would be more idiomatic.

borrowd_groups/__init__.py
ajk8, line 1: Same as above, can this be moved to `borrowd_groups/exceptions.py`. On top of generally being idiomatic, it helps to avoid circular dependencies as things get more complex.

borrowd_groups/models.py
ajk8: It's rather brilliant, really.

borrowd_groups/models.py
```diff
@@ -65,18 +101,43 @@ class BorrowdGroup(Group, GuardianGroupMixin):  # type: ignore[misc]
...
-            group=self,
-            trust_level=trust_level,
-        )
+        membership: Membership
+        # TODO: Should this simply be an .update_or_create() call?
+        # I'm opting _not_ at this point, because we should never
+        # logically be in a position where we're _trying_ to add
+        # a user to a Group they're already in: so if that happens,
+        # we should want to diagnose the situation that led us there.
+        # But, open to other opinions.
```
ajk8, line 126: In addition to the reason that you specified, I've seen `update_or_create` get buried in logic and cause massive performance issues when used at scale. I don't think we'd have that sort of issue here, but it seems best to avoid it in general and just be more explicit.

borrowd_groups/models.py
rjendoubi: Thanks! 7dc95b1 :heavy_check_mark:

borrowd/__init__.py
rjendoubi: Fair! d76d6c4 :heavy_check_mark:

borrowd_groups/__init__.py
rjendoubi: d76d6c4 :heavy_check_mark:

ajk8 (approved):
Thanks for the nit fixes! I'm good to roll with this 🚚

rjendoubi:
Hey @bdebut01 :wave: Since you requested changes the first time around, your specific re-review is needed to move to merge :pray: (I also have the "Dismiss Review" option, but that feels super rude and not the intention here :joy:)
Thanks!

bdebut01 (approved):

PR #43, opened 2025-05-14
Groups: object-level permissions base work

rjendoubi:
This is the prep for and implementation of object-level permissions for Groups - making it so not all Groups are visible to everyone on the Group listing page.

Not yet included:

* lock-down of Group detail page
* basic requirements for being logged-in
* moderator approval for new joiners
* ability to make someone a moderator
* ability to change member status
* ability to remove a member from the Group

Includes:

*    Extra tests verifying Item access on Group deletion.
    
    No additional permission management code required here, since Item
    viewing permissions are via the Group itself (except for the owner);
    so things "just work" when the Group is deleted, i.e., items to which
    one only had access by virtue of the deleted group are no longer
    accessible.

*    Fix test to raise when removing User not in Group.

*    Simplify impl for no re-adding same User to Group.
    
    Previous implementation created issues for additional checks to be
    added in subsequent commits. New implementation is more obvious
    anyhow.

*    Groups: require at least one Moderator at all times.
    
    Requires checks when removing members _and_ when updating Memberships
    to remove `is_moderator` status.

*    Add record-level permissions for Groups.

*    Apply record-level permissions for Groups.
    
    Plus tests.

*    Groups: listing page filtering and security.
    
    Basic `django-filter`-based filtering of the listing page, and
    `django-guardian` for security. See previous updates for Items
    for a bit more detail.


borrowd_groups/signals.py
```diff
@@ -35,6 +39,32 @@ def set_moderator_on_group_creation(
     )
 
 
+def _check_for_remaining_moderators(
```
bdebut01, line 42: how do we feel about `_raise_if_last_moderator` instead. Was stewing on how to make clear the relationship between the passed `user` and group. And function only raises an exception doesn't return anything. All that said, defer to you on what is most pythonic/django-y!

templates/groups/group_list.html
```diff
@@ -2,6 +2,24 @@
 
 {% block head_title %}Groups{% endblock %}
 {% block groups_content %}
+<c-box>
```
bdebut01, line 5: Quick one just to match the `items-list` page for now 😉 , below code mimics item-list html by wrapping overall in div to space the two blocks, converts list header to h2, and moves that 'Create New' button to up top!
```
{% extends "groups/layout.html" %}

{% block head_title %}Groups{% endblock %}
{% block groups_content %}
<div class="flex flex-col space-y-4">
    <c-box>
        <div class="flex justify-between items-center">
            <c-h1>Groups</c-h1>
            <c-button>
                <a href="{% url 'borrowd_groups:group-create' %}">Create New Group</a>
            </c-button>
        </div>
        <form method="get">
            {{ filter.form.as_p }}
            <input type="submit" value="Filter" class="bg-blue-500 text-white px-4 py-2 rounded">
            {% comment %}
            Only show the Clear button if filters are actually set.
            This looks a bit awkward in order to account for empty filters,
            i.e. `?category=&search=`
            TODO: Replace this with Alpine.js logic later
            when we're also using HTMX to avoid page reloads here.
            {% endcomment %}
            {% if not request.GET.values|join:"" == "" %}
            <c-button>
                <a href="{% url 'borrowd_groups:group-list' %}">Clear</a>
            </c-button>
            {% endif %}
        </form>
    </c-box>
    <c-box>
        <c-h2>Groups List</c-h2>
        <ul class="list-disc list-inside mb-4">
            {% for group in borrowdgroup_list %}
                <li>
                    <a class="hover:underline text-blue-800" href="{% url 'borrowd_groups:group-detail' group.pk %}">{{ group.name }}</a>
                </li>
                {% empty %}
                    <p>No groups found.</p>
            {% endfor %}
        </ul>
    </c-box>
</div>
{% endblock %}
```

image

bdebut01 (changes requested):
Minor suggestions

borrowd_groups/signals.py
rjendoubi: I like clear naming and I like this suggestion! Done in 05a06b67 :+1:

templates/groups/group_list.html
rjendoubi: Thanks! Done in 6b4cb18d9.

Also FYI, I read somewhere that it's possible to make "suggested changes" or something while doing PRs? Then at least your huge fixes on the templates would be credited to you :joy:

borrowd_groups/signals.py
```diff
@@ -35,18 +39,47 @@ def set_moderator_on_group_creation(
...
+        and not (
+            isinstance(origin, QuerySet) and isinstance(origin.model, BorrowdGroup)
+        )
+    ):
+        other_moderators = Membership.objects.filter(
+            group=group, is_moderator=True
+        ).exclude(user=user)
+
+        if not other_moderators.exists():
+            # If there are no other moderators, assign the owner as a moderator
```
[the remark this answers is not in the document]
rjendoubi: Good catch; fixed in 3fe17d9.

borrowd_groups/signals.py
```diff
@@ -59,3 +92,69 @@ def refresh_item_permissions_on_membership_update(
...
+    when their membership is deleted.
+    """
+    membership = instance
+    user: BorrowdUser = membership.user  # type: ignore[assignment]
+    group: BorrowdGroup = membership.group  # type: ignore[assignment]
+
+    #
+    # Check the group will not be left without a Moderator
+    #
+    _raise_if_last_moderator(user, group, **kwargs)
```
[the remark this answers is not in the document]
rjendoubi: :+1: 3fe17d9 adds a message to the exception.

borrowd_groups/signals.py
```diff
@@ -59,3 +92,69 @@ def refresh_item_permissions_on_membership_update(
...
+    If not, check if the group has any other moderators.
+    If not, raise a ModeratorRequiredException.
+    """
+    membership = instance
+    user: BorrowdUser = membership.user  # type: ignore[assignment]
+    group: BorrowdGroup = membership.group  # type: ignore[assignment]
+
+    # Check if the user is being added as a moderator
+    if not membership.is_moderator:
+        _raise_if_last_moderator(user, group, **kwargs)
```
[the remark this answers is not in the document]
rjendoubi: 👍 [3fe17d9](https://github.com/borrowd/borrowd/pull/43/commits/3fe17d9f0e6b02614de49bcbe2aa6083493ad653) adds a message to the exception.

ajk8 (approved):

krh-91 (approved):

bdebut01 (approved):

PR #55, opened 2025-05-21
33 items browser UI v1

bdebut01:
First pass at item browsing layout. 

Note: all kinds of `todos` in there. This PR is more focused on incremental improvements to our layouts, so in this case: very basic unordered list → componentized boxes w/ image displays. 

image


ajk8 (approved):
There are a lot of todos in here, but I think that's fine for now -- let's get this stuff rolling quickly and then we have something to iterate on. My favorite part of this pr is "Owned by donkey_teeth" 🫏 🦷 😆

PR #56, opened 2025-05-22
platform.sh prod setup

krh-91:
Support for platform.sh hosting
---
### Key takeaways:
- platform.sh gives us the flexibility to target multiple cloud providers with the same IaC definitions 
- platform.sh provides green data for the various hosting options (e.g. exact gCO2eq/kWh output for comparison between regions and providers)
- automatic git integrations for deploying to multiple environments (if eventually desired)
- provides a user-friendly dashboard for health, metrics, service mapping, etc across environments

### Hosting decisions (for discussion/review):
- region - an "Energy Grid Carbon Intensity" metric is provided on project creation when choosing the hosting region. In the US, Azure West (us-3) has by far the best performance in this metric although other considerations are relevant. Hosting region is locked in once the project is created, but since everything is IaC right now, it should be relatively easy to spin up a new project and export/import the db if we change our minds down the line. It doesn't look like multi-region support/sharding type behavior is available OOTB, but that would be a long ways away anyways.
- web server - gunicorn, since we don't need the async/web sockets support of daphne or uvi/hypercorn
- db - PostgreSQL. Open-source, feature rich, good django support.
- media/CDN - storing/serving uploaded files using a local container directory/mount until CDN integration work is done
- environments - this is primarily to support a prod environment, there will be some small additional work (noted in code comments) to spin up staging/cert environments if desired

Edited Copilot summary below
---
This pull request introduces significant changes to integrate the project with Platform.sh, including environment configuration, deployment settings, and dependency updates.

### Platform.sh Integration:

* Added `.env.platform` file to define non-sensitive environment variables for Platform.sh environments.
* Configured `.platform.app.yaml` for application deployment, including container settings, service relationships, web server settings, and deployment hooks (e.g., static asset collection and database migrations).
* Added `.platform/routes.yaml` to define domain-level URL routing
* Added `.platform/services.yaml` to configure PostgreSQL db service

### Django Configuration Updates:

* Updated `borrowd/config/base.py` to flexibly load Django secret key for different hosting environments.
* Enhanced `borrowd/config/env.py` to conditionally load `.env` files, supporting environments where `.env` files are unnecessary.
* Added `borrowd/config/prod/django.py` with Platform.sh-specific settings, including database configuration, static file paths, and secret key handling.

### Dependency Updates:

* Updated `pyproject.toml` to include `psycopg[binary]` for PostgreSQL support and `gunicorn` as the application server. Also updated `django-guardian` to version 3.0.0 and removed its custom source configuration.

### UI Improvement:

* Replaced the `Cancel` button in `templates/users/profile_form.html` with a `c-button-nav` component for improved semantic structure and maintainability.


.platform/services.yaml
```diff
@@ -0,0 +1,10 @@
+# The name given to the PostgreSQL service (lowercase alphanumeric only).
+db:
+    # The type of your service (postgresql), which uses the format
+    # 'type:version'. Be sure to consult the PostgreSQL documentation
+    # (https://docs.platform.sh/add-services/postgresql.html#supported-versions)
+    # when choosing a version. If you specify a version number which is not available,
+    # the CLI will return an error.
+    type: postgresql:17
+    # The disk attribute is the size of the persistent disk (in MB) allocated to the service.
+    disk: 1024
```
ajk8, line 10: We will want this to be larger for a production deployment. Maybe 5G to start?

ajk8 (commented):
This looks amazing. Can't wait to get it up and running!

.platform/services.yaml
krh-91: Think we will need to keep this at 1024 for now with the current platform.sh plan.
"E: Error: Resources exceeding plan limit; disk: 7168.00MB > 5120.00MB; try removing a service, or add more storage to your plan"

ajk8: Hm. Can we raise it to 2? 1G is really small. If we have to stay under that limit then fine, though I have a question: can we raise the disk size without an outage?

krh-91: Bumped it up to 2G. To change the disk size and avoid an outage we can do a blue-green deployment with a staging environment. Platform.sh supports environment promotions so should be straightforward.

PR #57, opened 2025-05-22
Borrowing flows: MVP UI integration

rjendoubi:
Not yet implemented:

* Full testing for e2e workflow (but it does work!)
* More mitigation for edge cases

PR includes:

* Borrowing flow: small cleanups
    
    * Standardize on "item_actions", not "action_options".
    * Add "CANCELLED" to exclusion for finding "current transaction"
    * Normalize ItemAction enum capitalization
    
    Re the last point: Turns out, if you define an enum like this:
    
        class Foo(TextChoices):
            MY_BAR = "my_bar", "My Bar"
    
    Then you have to do lookups in one of the following ways:
    
        Foo("my_bar")
        # ...or ...
        Foo["MY_BAR"]
    
    Using the wrong capitalization with the wrong brackets (function
    call vs dict member lookup) causes problems!
    
    Hence, standardizing at this point to both name and value as
    uppercase, to reduce opportunities for self-sabotage later.

* Add HTMX plumbing.
    
    Pretty minimal, ey? :)
    
    Although soon we should get this, Tailwind and Alpine bundling locally
    via `django-vite`. If for no other reason than so that I can keep
    hacking locally on trains with no WiFi.

* Borrowing flows: item action buttons.
    
    * Item action buttons component to an /items/ subdir
    * Add in HTMX attributes
    * Add component to item detail template

### DIFF
--- a/templates/items/item_detail.html
+++ b/templates/items/item_detail.html
@@ -5,6 +5,7 @@
 <c-box>
     <c-h1>{{ object.name }}</c-h1>
     <p>{{ object.description }}</p>
+    <c-items.action-buttons />
 
     <c-gallery>
     {% for photo in view.object.photos.all %}
--- a/borrowd_items/models.py
+++ b/borrowd_items/models.py
@@ -29,14 +29,14 @@ class ItemAction(TextChoices):
     user when viewing an Item.
     """
 
-    REQUEST_ITEM = "request_item", "Request Item"
-    ACCEPT_REQUEST = "accept_request", "Accept Request"
-    REJECT_REQUEST = "reject_request", "Reject Request"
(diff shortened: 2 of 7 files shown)


templates/items/item_detail.html
```diff
@@ -5,6 +5,7 @@
 <c-box>
     <c-h1>{{ object.name }}</c-h1>
     <p>{{ object.description }}</p>
+    <c-items.action-buttons />
```
bdebut01, line 8: lovely, was just thinkin after my last PR that we should start subdirectoring at this point 👍

bdebut01 (approved):

krh-91 (approved):
Looks good. Cool to see the main UI flow coming together!

PR #71, opened 2025-05-28
Group invites approval: pre-work

rjendoubi:
As flagged on Slack, this is a largeish PR with a pretty fundamental change in the way we do group-based permissions.

Video explanation: https://youtu.be/_7wiJDuYacU

Note this PR doesn't _yet_ actually implement any new functionality - it's just moving things around in preparation to be _able_ to provide that new functionality (group invite moderator approvals). The only "achievement" after this PR is that all the tests (should!?) still pass :sweat_smile: I ran it through a manual process of creating one group, one item, two users and doing a borrowing flow as well - grateful for any other ways you can think of to kick the tyres :muscle: 

Includes:

* `membership_requires_approval=False` for all existing tests.
    
    So that the whole world doesn't blow up as we implement the
    approval flow checks and logic.

* New Group Membership statuses, conditional default.
    
    Groundwork for respecting `moderator_approval_required` setting.

* Group model separation: BorrowdGroup vs auth.models.Group.
    
    This is more groundwork for implementing moderator approval for group
    memberships. To do so, this commit separates BorrowdGroup from
    auth.models.Group, i.e. separating the concerns of our domain-specific
    Group membership on the one hand from access permissions on the other.
    
    BorrowdGroup used to be a direct subclass of auth.models.Group. Now this
    is no longer the case. Instead, BorrowdGroup is its own, standalone
    model (a regular subclass of django.db.models.Model), and whenever a new
    BorrowdGroup is created, a "corresponding" permissions group is created
    in parallel. This gives us the ability for a BorrowdUser to be "in" a
    BorrowdGroup, from a domain / "user-facing" perspective, *without*
    necessarily being part of the corresponding permissions group - which is
    exactly the situation we need when a user's group membership is
    "pending", or certain other non-"active" statuses.
    
    The tests we had in place for the existing group / item permission
    functionality has enabled these changes to be made with a good degree of
    confidence - some light manual testing also seemed fine.
    
    Note that, since this a pretty fundamental change in the core auth
    models, there is NO MIGRATION PATH to this new state of affairs: you
    will have to delete your database and start from scratch. This has also
    given us the opportunity to "compact" several migrations for BorrowdUser
    and BorrowdGroup, recreating them from scratch at this point.


borrowd_groups/signals.py
```diff
@@ -12,6 +13,17 @@
...
+) -> None:
+    # Only run this logic on instance creation
+    if not created:
+        return
+
+    Group.objects.create(name=instance.name)
+
+
 @receiver(post_save, sender=BorrowdGroup)
 def set_moderator_on_group_creation(
```
ajk8, line 28: Is there any risk that these two things run out of order? If so, should they be combined into a single routine?

rjendoubi: So, my reason for opting for using Signals in general is because the logic is _"more"_ guaranteed to be applied.

Like, we could have a `create_group` service function somewhere which would incorporate our own logic of creating both the BorrowdGroup and the associated permissions group. But, there would be nothing (apart from fallible human PR review) stopping someone in some part of the code grabbing the main BorrowdGroup model and doing a basic  `BorrrowdGroup.objects.create()`, which can happen _anywhere_.

Whereas if you use Signals, you're wiring up that logic to "always" run whenever a new instance of such-and-such model is created (I'm caveating my absolute statements here to imply "give-or-take exceptions, asteroid strikes etc." but you know what I mean).

So to your question of, is there a risk of them running out of order: I can't think of one. It doesn't seem to me to matter much whether it's handled in a `post_save` handler (as here) or a `pre_save` handler, the important thing is they're hitched together in some reliable way.

ajk8 (approved):
Ship it!

PR #72, opened 2025-05-28
Configure prod email proxy

krh-91:
Configures prod to use the basic SMTP proxy offered by Platform.sh. This is a non-guaranteed, best-effort service so this may need to be improved in the future to something like SendGrid, but should be enough for now.

### DIFF
--- a/borrowd/config/prod/django.py
+++ b/borrowd/config/prod/django.py
@@ -59,9 +59,11 @@ if env("PLATFORM_APPLICATION_NAME") is not None:
     EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"
     EMAIL_HOST = env("PLATFORM_SMTP_HOST", default=None)
     EMAIL_PORT = 25
+    DEFAULT_FROM_EMAIL = "noreply@borrowd.org"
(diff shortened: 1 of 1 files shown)


borrowd/config/prod/django.py
```diff
@@ -56,6 +56,13 @@ def decode(variable: str):  # type: ignore
     # This should already be set correctly by the base config, but including this as insurance
     SECRET_KEY = env("PLATFORM_PROJECT_ENTROPY")
 
+    EMAIL_BACKEND = "django.core.mail.backends.smtp.EmailBackend"
+    EMAIL_HOST = env("PLATFORM_SMTP_HOST", default=None)
+    EMAIL_PORT = 25
+    EMAIL_USE_TLS = False
+    EMAIL_USE_SSL = False
```
ajk8, line 63: Why are these `False`? Or, perhaps a better question would be "when would/should these be set to `True`?"

krh-91: No luck using TLS with other ports, leaving a comment in the config explaining the situation

ajk8 (approved):

PR #73, opened 2025-05-29
Email group invite

krh-91:
Adding support for sending group invites via email/notifications system by adding an email input/form to the group invites page and support form submission group invites view to send emails. The view supports sending invites to multiple email addresses, but the template currently only shows one email input for now. Successful sharing reloads the page with a success message (as opposed to redirecting back to the group details page).

There aren't many checks right now on sending the emails (e.g. is the email address already associated with a user in the group? have they recently been sent an email? etc..). I was debating create a separate model to track sent invitations and send the email as a post-save signal, but that seemed like overkill right now and probably overly complex.

Also, added a debug launch configuration for VS Code users.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,16 @@
+{
+    "configurations": [
+        {
+            "args": [
+                "runserver"
+            ],
+            "autoStartBrowser": false,
+            "django": true,
+            "name": "Python Debugger: Django",
+            "program": "${workspaceFolder}/manage.py",
+            "request": "launch",
+            "type": "debugpy"
+        }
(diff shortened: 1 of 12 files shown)


ajk8:
> There aren't many checks right now on sending the emails (e.g. is the email address already associated with a user in the group? have they recently been sent an email? etc..). I was debating create a separate model to track sent invitations and send the email as a post-save signal, but that seemed like overkill right now and probably overly complex.

This seems important. A: You're right on with the checks and B: tracking invites could be very useful for metrics in the early-going. I would support holding this until we have the separate model and some more flow safety.

PR #75, opened 2025-06-02
Integrate `vite` for hot module reloading and packaging.

rjendoubi:
[`vite`](https://vite.dev) is the current hotness in frontend build tooling. Hooray for anything written in Rust - efficiency is green :)

This commit _looks_ like it's doing a lot at once, but most of these changes are just updates to various dependency metadata files, and bringing in some mostly-default configs for the new tooling.

The biggest change here is that we now require `node` and `npm`, to avail ourselves of the `vite` goodness.

This commit basically entails:

* Add `django-vite` as a dependency
* Remove `django_browser_reload` as a depenency, as the `vite` dev server is now performing this function and more
* Add tailwindcss vite plugin
* Add super simple main.js as entrypoint for vite bundling process
* Add super simple main.css to pull in local tailwind
* Update base templates:
  - No more tailwind from CDNs! Use locally provide from vite
  - Add vite template tags, enabling hot module replacement
* Updates to .gitignore
* Updates to CONTRIBUTING.md, including setup for `nvm` -> `node` -> `vite`

Not tackled yet:

* Converting other deps from CDN to locally bundled

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -5,12 +5,22 @@
 This project uses [`uv`](https://docs.astral.sh/uv/) and
 [`pre-commit`](https://pre-commit.com/).
 
-### 1. Install `uv` and tools
+### 1. Install `uv`, `npm` and other tools
+
+Recommended: install `npm` via
+[`nvm`, Node Version Manager](https://github.com/nvm-sh/nvm?tab=readme-ov-file#installing-and-updating):
 
 ```
+# Python stuff
 curl -LsSf https://astral.sh/uv/install.sh | sh
 uv tool install ruff
 uv tool install pre-commit
+
+# Javascript stuff
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.3/install.sh | bash
+# Activate nvm by opening a new terminal or sourcing ~/.nvm/nvm.sh; then:
+nvm install stable
+nvm use stable  # after this, `which npm` should find a working npm version
 ```
 
 ### 2. Clone repo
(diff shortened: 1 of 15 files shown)


rjendoubi:
The primary benefit is smaller download sizes for end-users: as opposed to downloading the entirety of Tailwind, they get an optimized bundle containing only the CSS/JS that our app uses. Pretty sure it does the same for JS packages as well (Alpine + HTMX in our case).

Also allows for offline development (which maybe no other than me cares about :man_shrugging:) and was the most straightforward way I could get editor support for Tailwind, but can maybe figure something else out in time.

But the main reason is optimized CSS/JS bundles - faster app experience, good for users with lower-end hardware, and (in principle, in a way so marginal that I wouldn't know how to measure it) "greener" :slightly_smiling_face:

ajk8 (approved):
Thanks for the explanation. Ship it!

CONTRIBUTING.md
```diff
@@ -37,18 +63,39 @@ Django project directory and the various sibling Django app dirs
...
 ```bash
 cp .env.example .env
 ```
 
-### 5. Django stuff
+The example file contains default values which are appropriate for
+local development, but you can tweak the values in your own `.env`
+to your liking.
+
+### 5. Running the app
```
bdebut01, line 76: RE the latest change you made to _opt in_ to running front/back separately--I'd add a few lines in here describing that process (changing that local env variable)

bdebut01 (changes requested):

rjendoubi:
Re-based to `main` :+1:

PR #86, opened 2025-06-03
Add my lends and my borrows to profile page

krh-91:
Rendering pending borrowing requests (to and from) in addition to borrowed and owned items in profile page

Notes
- fixing python version to 3.13 to avoid mypy complaints about not importing from 'typing_extensions' package
- add vs code launch config
- cleaning up some mypy issues around user and profile typing
- add pending transactions and borrowed items query methods to Transaction model
- creating shared item grid component to handle items and transactions (showing items involved in borrow transactions)
- reuse item-grid on item listing and profile page
- refactoring item-box to take direct parameters as opposed to an item object parameter to support rendering items from transactions. django-cotton does not seem to allow passing in related models without uglier hacks (the related models are converted to their string representations instead of full objects)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -11,7 +11,7 @@ repos:
     rev: v1.15.0
     hooks:
     -   id: mypy
-        args: [--strict]
+        args: [--strict, --python-version=3.13]
         additional_dependencies: [
             pytest==8.3.3,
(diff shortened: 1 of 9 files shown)


ajk8 (approved):
I'm not in love with the `@staticmethod` style, but I realize it's idiomatic in django. Carry on!

PR #89, opened 2025-06-04
Item grid/box styling updates

krh-91:
Styling updates for item grid/boxes
- remove gray background
- add ellipsized description
- add user avatar thumbnail/icon
- replaced racoon profile image fallback with avatar icon
- add time icon
- increase image height

I pulled the specific icons needed from lucide.dev until we decide on an icon library/handling


borrowd_web/templatetags/label_colors.py
rjendoubi, line 1: We have some Borrow'd theme-specific colors curated for us by the brand team, as shown around the middle of [this Figma board](https://www.figma.com/design/slbfIhZzOWEr1e5qfGrf9s/Borrow-d-Product?node-id=0-1). These have been made available [in the main base.html here](https://github.com/borrowd/borrowd/commit/392017473454b3a58f8262bb66096c0b3bf737d7#diff-690b81fad8df2a1f1ce37c846641d9247e3472d244f17039d2910a5ed0c98d5eR19-R36). `borrowd-indigo`, `borrowd-fern`, `borrowd-honey` and `borrowd-plum` are basically theme-specific blue, green, yellow and red colors respectively. Each comes in a `-300` and a `-600` variant.

Consider using the theme colors here in place of the Tailwind default ones?

static/icons/circle-user-round.svg
rjendoubi, line 1: We have been given a set of official icons from the brand team. [Here's a link to the "account_circle" icon](https://www.figma.com/design/slbfIhZzOWEr1e5qfGrf9s/Borrow-d-Product?node-id=6-83&m=dev) (you might have to zoom wayyyy in :smile:). It's very similar to this, but can we use the one from brand instead please for consistency?

static/icons/clock-4.svg
rjendoubi, line 1: Likewise for this one - brand hasn't given us a clock one yet - we can ask for it, but also thinking, based on the purpose for which we're using this icon, maybe [this handshake icon](https://www.figma.com/design/slbfIhZzOWEr1e5qfGrf9s/Borrow-d-Product?node-id=6-42&m=dev) could do in the meantime?

templates/components/items/item_box.html
```diff
@@ -2,7 +2,7 @@
 {% load label_colors %}
 <a href="{% url 'item-detail' pk %}" class="block no-underline">
     {% with box_size=box_size|default:'md' %} {# Define box_size with a default #}
-    <div class="bg-gray-100 shadow-md rounded-lg hover:-translate-y-2 transition-all duration-300 ease-in-out
```
rjendoubi, line 5: FWIW, I kinda liked this grey background, as it provided a visual boundary of the bottom of the card, where the edges aren't defined by the image itself.
Before:
list-main
After:
list-79

But I expect we're gna receive a lot more design guidance in general in the near future, so no point quibbling over it at this stage :slightly_smiling_face:

templates/components/items/item_box.html
```diff
@@ -13,22 +13,23 @@
...
                 </div>
             {% endif %}
-            <div class="flex flex-col py-4 px-3">
+            <div class="flex flex-col py-2 px-3x">
                 <!-- todo: deal with super long names -->
-                <div class="text-lg font-bold">{{name}}</div>
-                <div class="text-sm text-gray-500">Owned by {{ owner }}</div>
+                <div class="text-md font-bold">{{name}}</div>
+                <div class="text-sm/5 text-ellipsis overflow-hidden text-gray-600 line-clamp-2">{{description}}</div>
+                {% comment %} <div class="text-sm text-gray-500">Owned by {{ owner }}</div> {% endcomment %}
```
rjendoubi, line 26: Was it intended to remove the "Owned by" pill? If so, can we remove the code entirely please - unless there's a specific reason to keep it here commented?

templates/components/items/item_box.html
```diff
@@ -13,22 +13,23 @@
...
             </div>
             <!-- todo: chip component -->
-            <div class="absolute bottom-0 right-0 m-2">
+            {% comment %} <div class="absolute bottom-0 right-0 m-2">
                 <!-- todo: request/borrow/action flow -->
                 <div class="bg-green-400 rounded-2xl px-2 text-sm">{{ status }}</div>
-            </div>
+            </div> {% endcomment %}
```
rjendoubi, line 32: Again - commented code lying around makes me antsy. Keeping old iterations of stuff is what `git`'s meant to do for us :smile:

templates/components/items/item_grid.html
```diff
@@ -13,33 +13,47 @@
...
+                    {{ tx.created_at|days_since }}
+                {% elif label_type == 'tx_owner' %}
+                    <c-avatar src="{{tx.party1.profile.profile_pic}}" class="inline !h-7 !w-7" />
+                    From {{ tx.party1.username }}
+                {% elif label_type == 'tx_borrower' %}
+                    <c-avatar src="{{tx.party2.profile.profile_pic}}" class="inline !h-7 !w-7" />
+                    By {{ tx.party2.username }}
+                {% endif %}
+            </c-slot>
+            {% endif %}
```
rjendoubi, line 55: It feels a little strange for this conditional logic to live in the `item_grid` - would it not make more sense to push the `label_type` down into the `item_box` component itself and have the logic there?

Doing it that way would make it so we can leverage the `label_type` logic even when `item_box` cards are not appearing within an `item_grid` - maybe not happening right now, but could do later? Plus it just feels like that label is "part of the item card", so I'd encapsulate it in there.

rjendoubi (approved):
Looking good! No massively substantial feedback - just some suggestions to use the standardized stuff we have from the brand team, and or two other small things :+1: 

Thanks a lot!

templates/components/items/item_box.html
krh-91: Yes, I believe we agreed that not showing the item owner would avoid situations where requests are rejected and users end up having a negative interaction with each other. I left a new comment calling this out for future devs looking at this component and wondering why its not present :smile:

templates/components/items/item_box.html
krh-91: And sorry for the commenting, it is a habit from personal projects where I tend to flip between design iterations often when I'm undecided on something :blush:

templates/components/items/item_grid.html
krh-91: I agree generally speaking and was initially going to take that approach, but this becomes problematic if the label contains data that lives outside the `Item` model. In this case, the label is showing data that is only available at the `Transaction` model level (involved parties, transaction status, etc.).

The `items-grid` component is also doing a bit of double duty here by handling transactions as well as pure items, but duplicating this code to have something like `transactions-grid/box` seemed like it would just create a lot of headaches trying to keep the design in sync.

Not sure if there would have been a cleaner workaround here so I am all ears.

krh-91:
I implemented most of the feedback and left more information about why I made certain decisions. Thank you @rjendoubi for the thorough review!

PR #106, opened 2025-07-03
96 item detail borrowing flow UI revamp

bdebut01:


templates/items/item_detail.html
```diff
@@ -2,20 +2,52 @@
...
-    <c-button-nav url="{% url 'item-edit' object.pk %}">Edit</c-button-nav>
-    <c-button-nav url="{% url 'item-delete' object.pk %}">Delete</c-button-nav>
+        <div class="flex flex-col bg-gray-100 p-4 rounded-lg mt-6 border-2 border-gray-300">
+            <c-h2>Description</c-h2>
+            <p>{{ object.description }}</p>
+        </div>
+    </div>
 </c-box>
+
+<!-- REMOVE, JUST FOR DEBUGGING -->
```
bdebut01, line 40: for the life of me couldn't get a simple log of the values getting passed in from the backend (in React world: `console.log(props)` basically) so here is the temporary way i made it happen

borrowd_items/views.py
```diff
@@ -122,6 +126,8 @@ def get_context_data(self, **kwargs: str) -> dict[str, Any]:
         context = super().get_context_data(**kwargs)
         user: BorrowdUser = self.request.user  # type: ignore[assignment]
         context["item_actions"] = self.object.get_actions_for(user=user)
+        context["current_borrower"] = self.object.get_current_borrower()
+        context["requesting_user"] = self.object.get_requesting_user()
```
krh-91, line 130: Is there any way this logic can be removed in favor of adding it to the `get_actions_for` method? At a glance, it seems like all the information should already be available to that method. 

This also adds 2 extra DB calls to view an item which is not ideal.

templates/components/button_nav.html
```diff
@@ -1,5 +1,10 @@
 <a href="{{ url }}"
-  class="inline-block w-fit text-white font-medium py-2 px-4 rounded
-    bg-blue-600 hover:bg-blue-700 {{class}}">
+  class="inline-block w-fit font-medium py-2 px-4 rounded transition-colors duration-200
+    {% if variant == 'outlined' %}
+      border-2 border-blue-600 text-blue-600 hover:bg-blue-50 hover:border-blue-700 hover:text-blue-700
+    {% else %}
+      bg-blue-600 text-white hover:bg-blue-700
```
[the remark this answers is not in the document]
bdebut01: Eh, wound up removing the outlined variant. And applied templatetags refactor here, thanks

.platform.app.yaml
bdebut01, line 1: Github confused by rebase, this is [already on](https://github.com/borrowd/borrowd/blob/main/.platform.app.yaml#L93-L110) `main`

borrowd/config/prod/django.py
```diff
@@ -45,10 +45,6 @@ def decode(variable: str):  # type: ignore
         ".platformsh.site",
```
bdebut01, line 45: [Already on](https://github.com/borrowd/borrowd/blob/main/borrowd/config/prod/django.py#L87-L116) `main`

PR #113, opened 2025-07-14
110 make adding a photo part of the item creation form

krh-91:


bdebut01:
If it's easy, I'd say let's support adding multiple images otherwise good w/ this!

PR #152, opened 2025-09-05
Items | Description field is multiline

bdebut01:
+ Fix range of mobile styling issues
+ Refactor a bit in backend so item update and create share a form base

## Before
image
image
image
image

## After
image
image
image
image


ajk8 (approved):
Ship it!

PR #163, opened 2025-10-13
ability to disable borrowd_beta

ajk8:
Added a flag allowing us to disable borrowd_beta locally (and, eventually, on the live site). In order to use it, add the following to your local .env file:

```
BORROWD_BETA_ENABLED=False
```


tomjones-4 (approved):
Looks good to me

PR #166, opened 2025-10-25
Add image compression for profile pics, items, and groups

tomjones-4:


borrowd_groups/models.py
```diff
@@ -69,8 +70,22 @@ class BorrowdGroup(Model):
...
+        upload_to="groups/logos/",
+        processors=[ResizeToFit(800, 800)],
+        format="JPEG",
+        options={"quality": 75},
+        null=True,
+        blank=True,
+    )
+    banner: ProcessedImageField = ProcessedImageField(
+        upload_to="groups/banners/",
+        processors=[ResizeToFit(800, 800)],
```
ajk8: Is the banner supposed to be square? I would think that is a usually a wide image.

borrowd_groups/models.py
```diff
@@ -69,8 +70,22 @@ class BorrowdGroup(Model):
     description: TextField[Never, Never] = TextField(
         max_length=500, blank=True, null=True
     )
-    logo: ImageField = ImageField(upload_to="groups/logos", null=True, blank=True)
-    banner: ImageField = ImageField(upload_to="groups/banners", null=True, blank=True)
+    logo: ProcessedImageField = ProcessedImageField(
+        upload_to="groups/logos/",
+        processors=[ResizeToFit(800, 800)],
```
ajk8: How did you come to settle on 800x800? Is that a best practice or just something that you found seems to work?

PR #181, opened 2025-11-11
disable beta wall functionality by default

ajk8:
The previous implementation for disabling the beta functionality via environment variable was not fully functional. This change does a few things:

1. Make the beta wall functionality disabled by default
2. Allow it to be turned on by environment variable
3. Separately, an environment variable has been set in staging which turns on the beta wall functionality


bdebut01 (changes requested):
Works locally on my machine! 

Two `md` updates if you don't mind...
1. Adding your `BORROWD_BETA_ENABLED` env variable to the `.env.example` file
2. Adding the same variable to the `CONTRIBUTING.md` under `Env vars` with description etc

Thanks!

PR #221, opened 2025-12-16
219 groups admins can remove group members

bdebut01:
## Moderator view (get confirm dialog on remove click)
image

## Kicked User's View of Group
image


ajk8 (approved):
Assuming that this correctly makes the removed user's disappear from the group's inventory, then I say ship it! I believe that the signals code will handle that on membership change without any changes, but it should be tested to make sure.

PR #234, opened 2025-12-30
added permissions checking for all current objects

### LINKED ISSUE #9: Lock down URLs for groups based on permissions
Users should only see Groups of which they are Members.

### LINKED ISSUE #10: Lock down URLs for items based on permissions
Users should only see Items in Groups where they are Members, and where the Owner of the Item has assigned it a Trust Level at or lower than the Trust Level they have via their Group Membership.

e.g. If User A has a Medium Trust Level with Group X, then any High Trust Items posted by User A will not be visible to other Members of the Group: only Medium or Low Trust Items will be visible.

ajk8:
Fixes #9 
Fixes #10 

This PR adds permission checking for all current endpoints. In general, it follows this pattern:

* if a request passes permission checks, return the appropriate view
* else if a user is not logged in, redirect to the login flow
* else return a 404 or 403, as appropriate to the request

When a 404 is returned, it uses a custom 404 page. This is ugly and needs UX help!

image

This PR also contains some miscellaneous under-the-hood fixes:

1. Slightly better logging support
2. Improved error handling on shared templates
3. Make permissions enum, rather than copy-paste strings

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -1,5 +1,4 @@
 # see CONTRIBUTING.md for descriptions of the different environment variables
 DJANGO_SETTINGS_MODULE=borrowd.config.dev.django
 DJANGO_SECRET_KEY=not-a-real-secret
-BORROWD_USE_LOCAL_BUNDLING=False
 BORROWD_BETA_ENABLED=False
--- a/borrowd/config/base.py
+++ b/borrowd/config/base.py
@@ -38,7 +38,6 @@ ALLOWED_HOSTS: list[str] = []
 # Borrowd settings
 #
 BORROWD_GROUP_INVITE_EXPIRY_SECONDS: int = 60 * 60 * 24 * 7  # 1 week
-BORROWD_USE_LOCAL_BUNDLING = env("BORROWD_USE_LOCAL_BUNDLING", default=False)
 
 # Application definition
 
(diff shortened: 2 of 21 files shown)


borrowd_groups/signals.py
vguzman812: Much better readability/maintainability here now. Nice 👍

borrowd_groups/tests/test_group_detail_view.py
```diff
@@ -71,13 +72,10 @@ def test_logged_in_non_member_cannot_view_detail_page(self) -> None:
...
         # Act
         #
-        response = GroupDetailView.as_view()(request, pk=group.pk)
-
-        #
-        #  Assert
-        #
-        # Forbidden error
-        self.assertEqual(response.status_code, 403)
+        # the Mixin raises a PermissionDenied which does not get caught in this test, but
```
vguzman812: Would something like getting the client and checking status code that way work? https://docs.djangoproject.com/en/5.2/topics/testing/tools/

ajk8: I was able to get the client working, but it causes another problem. When running the tests, they spit out the PermissionDenied traceback to the console which makes it hard to read the results. This is not a huge problem, but I don't know that it's any better than the current implementation.

borrowd_groups/views.py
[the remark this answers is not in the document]
ajk8: `GroupJoinView` is used for any authenticated user to join a group for which they have an invite link, so `LoginRequiredMixin` is appropriate.

`UpdateTrustLevelView` takes care of group membership check in the logic of the function. This could be refactored out, but I don't know that it's necessary.

`RemoveMemberView` also takes care of group membership (and moderator) permissions in the logic of the function.

vguzman812 (approved):

borrowd_groups/tests/test_group_detail_view.py
vguzman812: Let's leave it as is, then.

Issue #254, opened 2026-01-22
No search results returns "zero state" onboarding flow


ik-brwd:
When a user searches for an item that returns zero results, instead of seeing zero results, the user is shown the "add an item" zero state (as part of onboarding flow). 

This loses the search term box etc. 

See below: 

https://github.com/user-attachments/assets/a780da54-fe11-4e64-9a45-6a9f33107bdf

Issue #263, opened 2026-01-28
Updated Search Page


ik-brwd:
Minor amends to new search page: integrate DaisyUI components; integrate new Modal for categories; integrate new item cards. 

Includes: 
#216 
#228 

<img width="740" height="945" alt="Image" src="https://github.com/user-attachments/assets/7f5dff25-186f-438e-9f83-7a383aad72fe" />

Issue #228, opened 2025-12-19
Search flow UI: Homepage with items


sallynorthmore:
When there are items in `/items`, the search and filters components should be visible and layout should match the designs. 

Work might* include: 

- layout of search and filters components at top of page
- items list layout

* small caveat that some of this work may have been already completed in other tasks, in which case this ticket is to check all looks well

<img width="299" height="540" alt="Image" src="https://github.com/user-attachments/assets/8ace4dae-bf10-4229-bf3a-739effac1a51" />

Issue #216, opened 2025-12-14
Search Flow UI: Filters Modal


sallynorthmore:
Page: https://staging-5em2ouy-wrpvbokn22dsg.us-3.platformsh.site/items/

In scope:

- Clicking the "Filter by category" button from the search page should trigger the filter modal to open
- Filters tags should use daisy components where possible for hover/selected/unselected states
- Selected filters should appear with dark background 
- Update search/clear filters button group
- Update search click should close modal and update search results
- Clear filters  click should close modal and unselect all filters, updating search results 

<img width="228" height="638" alt="Image" src="https://github.com/user-attachments/assets/9a17198d-002e-4abb-9b0e-5a7271eeb65b" />

Figma: https://www.figma.com/design/wMliTL8KGBlUACk0d8fkZ3/Borrow-d---Mobile-App--mid-fidelity-?node-id=716-14575&t=fbjk4d6JhpxK3uIV-0

PR #279, opened 2026-02-03
Feat/216 filters modal

vguzman812:
# Summary
Adds a modal to select multiple categories for items. Slightly changes the search page UI to better align with [Figma designs.](https://www.figma.com/design/wMliTL8KGBlUACk0d8fkZ3/Borrow-d---Mobile-App--mid-fidelity-?node-id=716-14690&m=dev)
Needs to be incorporated with the item cards to finish the search page UI.
Addresses a bug in the base modal component that made the close "X" button sometimes not register. Also changed the modal to use Alpine JS `@click` rather than `onclick`. This will prevent global name mismatches (think `window.confirm()`) and allow functions in alpine.js to be passed rather than them being attached to the window object. 

# Issues addressed

**closes:**
#216

I believe this also closes:
#254
See no items found section in the media below.

**Partially addresses:**
#263 
#228

# Media:

### Item Creation

Screenshot 2026-02-03 at 10 12 48 PM

### Item search

Screenshot 2026-02-03 at 10 12 18 PM

### Item search: no items found (see URL in screenshot)

Screenshot 2026-02-03 at 10 13 59 PM

### item creation/editing

https://github.com/user-attachments/assets/1bb37517-7529-4e94-8336-cf6f243da709

### item search

https://github.com/user-attachments/assets/3d37fc1c-19bd-4af0-a134-d0952ebec892

### item search no items found

Screenshot 2026-02-03 at 9 55 49 PM

### DIFF
--- a/templates/items/layout.html
+++ b/templates/items/layout.html
@@ -1,7 +1,7 @@
 {% extends "layouts/default.html" %}
 
 {% block content %}
-    <div class="max-w-3xl">
+    <div class="w-full max-w-3xl self-start px-4">
     {% block items_content %}
     {% endblock %}
     </div>


templates/components/items/filter_form.html
vguzman812: @ajk8 idk how much Alpine JS or Django sanitizes comments for templating, and I'm not sure how important a security issue it is for my comments to be included here.
I'm also pretty tired, so I don't think I could look this up atm.  
The comments are almost exclusively for us (devs) so I know what the heck is happening in a week when I look back at this and inevitably think "who wrote this garbage and what does it do?"

templates/items/item_form.html
```diff
@@ -20,7 +20,113 @@
...
+                            multiple
+                            class="hidden">
+                        {% for choice in field.field.choices %}
+                        <option value="{{ choice.0 }}" {% if field.value and choice.0|stringformat:"s" in field.value %}selected{% endif %}>
+                            {{ choice.1 }}
+                        </option>
+                        {% endfor %}
+                    </select>
+
+                    {% include "components/items/category_filter_modal.html" with modal_id="item-category-modal" confirm_button_text="Filter categories" %}
```
bdebut01: especially for `item form` not the `filter`/`search` page, I think I'd vote for `Update categories` or `Select categories` or something. Because the user isn't filtering anything on that click, they're just saving their selected Categories.

templates/components/items/filter_form.html
```diff
@@ -1,42 +1,119 @@
...
-            i.e. `?categories=&search=`
-            TODO: Replace this with Alpine.js logic later
-            when we're also using HTMX to avoid page reloads here.
-            {% endcomment %}
-            {% if not request.GET.values|join:"" == "" %}
-                <c-button-nav url="{% url 'item-list' %}" class="btn btn-ghost">Clear</c-button-nav>
-            {% endif %}
-        </div>
+<form method="get" class="flex flex-col gap-4"
+    x-data="{
```
bdebut01: Given the number of identical/near-identical functions in this Alpine x-data vs. the `item_form` one, do you think it'd be possible or worth extracting into a reusable JS function?

bdebut01: Yup, suspected that'd be the case. Agreed to punt for now then! Thanks

PR #282, opened 2026-02-04
Feat/261 Add explanation screens

### LINKED ISSUE #261: Onboarding / explanation screens
Following account creation, new users should be served 3 static pages as product onboarding flow: 

The progress bar is a DaisyUI component we haven't integrated yet. (Needs to be done for these pages.)

[Figma file](https://www.figma.com/design/wMliTL8KGBlUACk0d8fkZ3/Borrow-d---Mobile-App--mid-fidelity-?node-id=88-209&p=f&t=0z1ZQpImltRQBZpD-0) (request access from ik@borrowd.org if required)

Image

kelvintran1605:
Closes #261 
## Summary
Added a 3-step onboarding explanation flow that users are redirected to after signing up.

Step 1:
Step 1
Step 2: 
Step 2
Step 3: 
Step 3

## Behavior
- Onboarding appears immediately after successful signup
- Users can navigate through the 3 explanation steps
- After completing the flow, users proceed to home page

### DIFF
--- a/borrowd_users/views.py
+++ b/borrowd_users/views.py
@@ -140,9 +140,9 @@ class CustomSignupView(CreateView[BorrowdUser, CustomSignupForm]):
             self.request, "Welcome! Your account has been created successfully."
         )
 
-        # Honor next parameter if it exists
-        if "next" in self.request.GET:
-            return redirect(self.request.GET["next"])
+        next_url = self.request.GET.get("next")
+        if next_url:
+            self.request.session["post_onboarding_redirect"] = next_url
 
         return redirect(self.success_url)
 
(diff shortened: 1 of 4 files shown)


bdebut01 (dismissed):
As discussed in Tech Sync, we'll want to if possible have users that recently registered via a Group Invite referral link to still go through these onboarding steps. 

Code looks great, zero notes there! Just let me know if you want more elaboration on above :D . And let me know if the above turns out to be a pain, don't think we want to make dramatic backend changes or anything for this.

bdebut01 (approved):
Thank you! Looks great!

PR #283, opened 2026-02-06
fix/253 : "remove-member" dropdown position

Pejjp:
**Problem**
The actions dropdown for a member could overflow/clipped when opened near the bottom of the Members panel (especially the last rows).

**Solution**

1. Added a boundary marker to the Members list container so the dropdown can compute available space relative to the panel (not the full window).
2. Implemented dynamic placement logic to open the dropdown above when there isn’t enough space below, otherwise open below.
3. Kept positioning anchored to the actions button (relative parent + absolute menu).

https://github.com/user-attachments/assets/d9a946cd-c804-4092-9eda-ece22d9321f8

### DIFF
--- a/templates/components/groups/member_row.html
+++ b/templates/components/groups/member_row.html
@@ -14,15 +14,50 @@
     </div>
     <!-- member actions -->
     {% if is_moderator and member.user_id != user.id %}
-    <div class="relative" x-data="{ open: false }">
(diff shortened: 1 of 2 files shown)


ajk8 (changes requested):
One change requested. Otherwise this looks find to me.

ajk8 (approved):

PR #295, opened 2026-02-16
Issue #269 - Public profile page (As seen by other users)

### LINKED ISSUE #269: Profile page - as seen by other users
We need to create a new page that captures the information from user profiles. 

Use case (for example): clicking someone's name in an item card to know who they are before responding to a request. 

Image

kelvintran1605:
Closes #269 

## What’s added

- Implemented dynamic public profile route: `/profile/<user_id>/`
- Added `public_profile_view` to fetch user by ID using `get_object_or_404`
- Created `public-profile.html` template
- Displayed:
  - Profile image (fallback to UI avatar if none)
  - Full name
  - Bio (with default fallback message)
  - Email

## Details

- Added new URL pattern in `borrowd_users/urls.py`
- Implemented view logic in `borrowd_users/views.py`
- Used Django template conditionals for image fallback
- Ensured safe user lookup using `get_object_or_404`

## Notes:
This only handles displaying public profile data. Linking the profile card to this route can be implemented separately if needed.

### DIFF
--- a/borrowd_users/views.py
+++ b/borrowd_users/views.py
@@ -12,20 +12,42 @@ from django.views.generic import (
     CreateView,
 )
 from django.shortcuts import get_object_or_404, render
+from django.http import Http404
 
 
 from borrowd_items.models import Item, ItemStatus, Transaction
+from django.http import HttpResponseForbidden
+from borrowd_groups.models import Membership
 
 from .forms import ChangePasswordForm, CustomSignupForm, ProfileUpdateForm
 from .models import BorrowdUser
 
+@login_required
 def public_profile_view(request: HttpRequest, user_id : int) -> HttpResponse:
     user = get_object_or_404(BorrowdUser, id=user_id)
     profile = user.profile
+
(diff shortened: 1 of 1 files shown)


borrowd_users/views.py
```diff
@@ -11,12 +11,21 @@
...
 from .forms import ChangePasswordForm, CustomSignupForm, ProfileUpdateForm
 from .models import BorrowdUser
 
+def public_profile_view(request: HttpRequest, user_id : int) -> HttpResponse:
+    user = get_object_or_404(BorrowdUser, id=user_id)
+    profile = user.profile
+    return render(request, "users/public-profile.html", {
+        "profile": profile,
+        "user_obj": user
+    })
```
ajk8: This view needs to be restricted to users who can view your items. That probable just means users that are in groups with you. Any group will do.

kelvintran1605: image
I've just pushed a new commit
Now it should return 404 if the users are not in the same group. Please let me know if this logic works !

PR #362, opened 2026-03-10
Fix/354 - Fixed standard not displayed and removed notifications

### LINKED ISSUE #354: Standard trust level is not displayyed in the item card
When the user updates the trust level from high to standard for an item, the trust level does not display a value in the item card.

Preconditions: User is logged in, user has an active item

1. Navigate to your item card with a high trust level
2. Click edit item and update trust level to standard
3. Check the item card again

Actual result: the trust level field displays no value

Image

Expected result: To display standard trust level

kelvintran1605:
Closes #354 
# Summary 
- Removed unnecessary notifications
- Fixed **Standard** trust level not displayed.
---
image
image


vguzman812:
This is good, and I know it aligns with what the current Figma design shows. However, there is an action that is lost by removing the action context here in the item detail page: the ability to cancel an item request after the owner has accepted the request.

The search/inventory item card view hides the cancel request action once the request has been accepted (i.e., when both `CANCEL_REQUEST` and `MARK_COLLECTED` are valid actions simultaneously). See`build_item_card_context` in `card_helpers.py:194-203`:
```python
def build_item_card_context(...)
...
    # Once the request is approved, the card only shows "Confirm picked up".
    # Cancel is still accessible on the item detail page.
    if (
        ItemAction.CANCEL_REQUEST in action_context.actions
        and ItemAction.MARK_COLLECTED in action_context.actions
    ):
        action_context = ItemActionContext(
            actions=tuple(
                a for a in action_context.actions if a != ItemAction.CANCEL_REQUEST
            ),
            status_text=action_context.status_text,
        )
...
```


So the flow that would be blocked: 
1. User requests an item 
2. Owner accepts the request
3. At this point, both CANCEL_REQUEST and MARK_COLLECTED are valid actions
4. The inventory card hides CANCEL_REQUEST
5. The only way for the borrower or lender to cancel at this stage is via the item detail page

If we remove actions from the item detail page, users will have no way to cancel a request that has already been accepted. The only way to move the flow forward would be for the item to be marked as collected.

I think a better solution would be to just have the actions be present in the item detail page still, but have them styled so it doesn't break the layout.

P.S. If there was some discussion that was had during the meeting today that I missed that covers this, feel free to ignore this request.

vguzman812 (dismissed):
Please keep the action buttons in the item detail page and just style them so they do not break the card.

ik-brwd:
See "Item details" section in [Ready for Development in Figma. ](https://www.figma.com/design/wMliTL8KGBlUACk0d8fkZ3/Borrow-d---Mobile-App--mid-fidelity-?node-id=697-12023)

PR #364, opened 2026-03-10
chore: add baseline gitattributes defaults

vguzman812:
As the number of people working on the project increases, it will inevitably come to pass that some will be using different operating systems. To prevent requiring all contributors to edit their git configs before contributing, I saw it fit to add a simple .gitattributes file to normalize line endings, among a few other things.


templates/components/groups/members_list.html
vguzman812: This file previously had CRLF line endings. It now has LF line endings. No other changes were made to it.

PR #369, opened 2026-03-11
311/ Feat/membership approval

kelvintran1605:
# Summary
- Add a moderator-only **Pending Approval (N)** section above the **Members** section on Group Detail, and hide it entirely when there are no pending requests.
- Render each pending membership row with avatar, clickable username, requested trust level, and **Approve / Deny** actions.
- Keep **pending memberships separate from active members**; the Members section should show ACTIVE members only.
- Implement **Approve** and **Deny** actions using the membership approval endpoints, update the UI after success, and show the required success/error toasts.
- Show a **Pending** badge on the Groups Index for users whose membership is pending.
- After a user requests to join a private group, redirect them to the groups list and show the pending state/toast.
- Prevent pending users from entering Group Detail from the Groups Index; show a pending-status toast instead.
- If a pending user revisits the invite link, show the pending request message instead of treating it like a new join.
---
This is the flow when I created two mock users and tested the approval flow.

https://github.com/user-attachments/assets/e4ee8dcd-3817-4507-8479-2b3d45a921ca

### DIFF
--- a/borrowd_groups/views.py
+++ b/borrowd_groups/views.py
@@ -9,12 +9,11 @@ from django.forms import ModelForm
 from django.http import (
     HttpRequest,
     HttpResponse,
+    HttpResponseBase,
     HttpResponsePermanentRedirect,
     HttpResponseRedirect,
 )
 from django.shortcuts import get_object_or_404, redirect, render
-from django.http import HttpRequest, HttpResponse, HttpResponseBase
-from django.shortcuts import redirect, render
 from django.urls import reverse, reverse_lazy
 from django.views.generic import CreateView, DeleteView, DetailView, UpdateView, View
 from django_filters.views import FilterView


borrowd_groups/models.py
```diff
@@ -152,7 +152,9 @@ def add_user(
                 (f"User '{user}' is already a member of group '{self}'")
             )
 
-        if self.membership_requires_approval:
+        if is_moderator:
+            default_status = MembershipStatus.ACTIVE
+        elif self.membership_requires_approval:
             default_status = MembershipStatus.PENDING
         else:
             default_status = MembershipStatus.ACTIVE
```
ajk8:
```suggestion
        if self.membership_requires_approval and not is_moderator:
            default_status = MembershipStatus.PENDING
        else:
            default_status = MembershipStatus.ACTIVE
```
Nit: this could be simplified

vguzman812 (approved):

PR #384, opened 2026-03-24
feat: added search term storage in table (#318)

### LINKED ISSUE #318: Store search terms in table
We are not capturing search terms users input, but this could be a valuable data insight.

ansonnchan:
Closes #318

**Summary**

This PR implements logging of user search terms for both items and groups. It stores **unique searches** in the `SearchTerm` model, deduplicating by user, target, and normalized term, and updates timestamps on repeated searches. An admin-only JSON export endpoint is also included.  

**Key changes**
- Added `SearchTerm` model with:
  - `term_raw` (original user input)
  - `term_normalized` (e.g. so "COMPUTER" and "computer" count as the same search)
  - `target` (items or groups)
  - `last_searched_at` (time of last search)
- Updated `ItemListView` and `GroupListView` to record searches
- Added `search_terms_export_view` for admin JSON export

Below is an example of what an admin user would see. Regular users would hit a 403 Forbidden if they tried accessing somehow.

Screenshot 2026-03-23 202600


borrowd_users/models.py
```diff
@@ -59,3 +63,89 @@ def profile_pic(self) -> str:
...
+    # Used for dedupe; lowercased + whitespace collapsed.
+    term_normalized: models.CharField[str, str] = models.CharField(max_length=200)
+
+    created_at: models.DateTimeField[Never, Never] = models.DateTimeField(
+        auto_now_add=True,
+    )
+    last_searched_at: models.DateTimeField[Never, Never] = models.DateTimeField(
+        default=timezone.now,
+    )
+
```
ajk8: This shouldn't be necessary. See explanation further down this file.

borrowd_users/models.py
```diff
@@ -59,3 +63,89 @@ def profile_pic(self) -> str:
...
+            )
+        ]
+        indexes = [
+            # Fast "latest searches per user per target" queries.
+            models.Index(fields=["user", "target", "-last_searched_at"]),
+            # Fast "latest searches by target" analytics queries.
+            models.Index(fields=["target", "-last_searched_at"]),
+        ]
+        ordering = ["-last_searched_at"]
+
```
ajk8: All of this would have to be removed or changed if we take my other suggestions.

PR #404, opened 2026-03-31
Same Group name available for different users

Pejjp:
## Title

Allow duplicate group names across different users

## Summary

This PR changes `BorrowdGroup` name uniqueness from global to per-user, so different users can now create groups with the same display name.

To keep the linked Django auth `Group` unique, the associated permissions group now uses an internal computed name based on the Borrowd group name and creator ID. The app also stops looking up auth groups by display name and instead uses the stable `perms_group` relation directly.

This PR also improves the create-group UX by catching duplicate names for the same user before the database write and surfacing the existing notification toast instead of only failing at save time.

## What Changed

- Removed the global unique constraint from `BorrowdGroup.name`
- Added a composite uniqueness constraint on `BorrowdGroup`:
  - `("name", "created_by")`
- Added a migration for the schema change
- Introduced `compute_per_group_unique_name(base_name, user_id)` for the linked auth `Group`
- Updated the `BorrowdGroup` post-save signal to:
  - create the auth group with a per-user unique internal name
  - keep that auth group name in sync on updates
- Updated membership and permission code to use `perms_group` directly instead of resolving `Group` by name
- Added targeted typing casts in signals to satisfy mypy around Django relation fields
- Added pre-save duplicate-name validation in `GroupCreateForm` for the current user
- Updated `GroupCreateView` to pass `request.user` into the form
- Reused the existing Django messages toast/banner to show a friendly error when a user tries to create a second group with the same name

## Plus
- Changed the group-form (create and edit) page header with the header component: This ensures design and behaviour consistency across all pages. 

## Why

Django auth groups still require globally unique names, but Borrowd groups now need user-scoped uniqueness. This keeps the user-facing group name flexible while preserving a reliable unique permissions group behind the scenes.

Adding the duplicate-name check in the form avoids relying on a database failure for normal user feedback and gives a cleaner error path in the UI.

## Testing

- Migration applies cleanly
- Creating groups with the same name for different users succeeds
- Creating duplicate groups with the same name for the same user is still rejected
- Duplicate-name rejection now happens during form validation before the insert
- The existing notification toast is shown for the duplicate-name case
- Membership and permission behavior continues to use the linked `perms_group`


vguzman812 (changes requested):
Tests need to pass for this to be approved. 
Additionally, I'm seeing that editing a group to a non-unique name that the same creator already uses for one of their groups triggers an integrity error.
I'm also seeing that new members of groups with the same name don't see items that they should. I.E., user A joins "group a" made by user B, and "group a" made by user C, and can only see items from one of them.

Pejjp:
Thank you for the feedback, I addressed the fixes for the bugs mentioned and added some unit tests to further validate.

vguzman812 (dismissed):
Hi, I'm still having several issues. Let's say there is "group A" that requires moderator approval before joining. User A is a part of group A and has a few items posted. User B tries to join group A and is now "pending" approval. However, user B can see all available items in group A, despite not being part of the group yet technically. User A cannot see the items of user B.
Additionally, altering past migrations is not a practice I think we should consider.

borrowd_users/migrations/0009_increase_bio_max_length.py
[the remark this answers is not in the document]
Pejjp: Restored the file

borrowd_groups/migrations/0010_alter_borrowdgroup_name_and_more.py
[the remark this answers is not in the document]
Pejjp: Pushed the new migrations

borrowd_groups/signals.py
```diff
@@ -128,7 +146,7 @@ def refresh_permissions_on_membership_update(
     #
     user = instance.user
     borrowd_group = instance.group
-    group = borrowd_group.perms_group
+    group = borrowd_group.perms_group  # type: ignore[attr-defined]
```
ajk8: These attr-defined issues should not be ignored. Or at least the ones below should not be ignored. If for some reason `perms_group` is not set, then when you try to add it to `user.groups`, you'll get an error.

borrowd_groups/signals.py
```diff
@@ -145,8 +145,8 @@ def refresh_permissions_on_membership_update(
     # Handle Item permissions
     #
     user = instance.user
-    borrowd_group = instance.group
-    group = borrowd_group.perms_group  # type: ignore[attr-defined]
+    borrowd_group: BorrowdGroup = instance.group  # type: ignore[assignment]
+    group = borrowd_group.perms_group
```
ajk8, line 150: Instead of changing the ignore statement, can you handle when the group is `None`?

borrowd_groups/signals.py
ajk8: This is where I would expect the code to blow up if group is `None` from the assignment above.

ajk8: It could also blow up here when group is `None` from the assignment above.

ajk8 (changes requested):
I think the issue with the type ignores still needs to be addressed.

ajk8 (approved):

chachan (approved):

Issue #351, opened 2026-03-08
[mobile] only digit keyboard opens when user is on enter sign-in cod creen


LenaLisovskaia:
When the user requests a sign-in code from their phone, then on enter sign-in screen, the keyboard has only digits available, but the code contains both letters and digits. 

<img width="590" height="1278" alt="Image" src="https://github.com/user-attachments/assets/570e9418-0052-4152-829f-b58a01f97bad" />

PR #412, opened 2026-04-01
login code generator function overide

Pejjp:
## Summary

This PR implements digit-only login codes for email authentication by overriding the default login code generator in allauth.

## Changes

- **borrowd/config/base.py**: Added `ACCOUNT_ADAPTER` to use the custom adapter and configured `ACCOUNT_LOGIN_BY_CODE_FORMAT` for 6-digit numeric codes.
- **borrowd_users/adapters.py** (new file): Created `BorrowdAccountAdapter` class with `generate_login_code()` method that generates a 6-character string using only digits.

## Motivation

Addresses issue #351 by ensuring login codes are numeric-only for better usability and security.

## Testing

- Verified the adapter generates 6-digit codes.
- Confirmed allauth uses the custom adapter via the config settings.

### DIFF
--- a/borrowd/config/base.py
+++ b/borrowd/config/base.py
@@ -138,6 +138,8 @@ AUTH_USER_MODEL = "borrowd_users.BorrowdUser"
 
 LOGIN_REDIRECT_URL = reverse_lazy("item-list")
 
+ACCOUNT_ADAPTER = "borrowd_users.adapters.BorrowdAccountAdapter"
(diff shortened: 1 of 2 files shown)


chachan:
understood, sorry I didn't realize code format changed too.
@Pejjp, I am noticing that when I access to the "Sign-In Code" page, it is outputting the whole HTML into the console. Is that expected?

PR #434, opened 2026-04-03
360 add missing lifecycle fields

### LINKED ISSUE #360: Add missing lifecycle fields to core models
The following models are missing lifecycle fields:

| Model | created_at | created_by | updated_at | updated_by | deleted_at | deleted_by |
| --- | --- | --- | --- | --- | --- | --- |
| `Item` | X | X | X | X | X | X |
| `ItemPhoto` | X | X | | | X | X |
| `Transaction` | | X | | | X | X |
| `BorrowdGroup` | | | | | X | X |
| `BorrowdUser` | X | X | X | X | X | X |
| `Profile` | | | X | X | | |

These fields all need to be add to the models. `BorrowdGroup` has examples of setting up the created_ and updated_ fields. The deleted_ fields should always be unset unless we have a delete action.

chachan:
This closes #360


vguzman812 (dismissed):
I'm getting that 80 tests fail. You seem to have made the migrations and updated the models, but not added the necessary fields to the records being created/updated.

borrowd_items/models.py
```diff
@@ -126,6 +128,45 @@ class Item(Model):
...
+        blank=True,
+        default=None,
+        help_text="Set when the record is soft-deleted. NULL means active.",
+    )
+    deleted_by: ForeignKey[BorrowdUser] = ForeignKey(
+        BorrowdUser,
+        null=True,
+        blank=True,
+        default=None,
+        on_delete=SET_NULL,
```
ajk8, line 166: Why is this one SET_NULL?

ajk8 (commented):
Migrations pass on my local now!

I see that updated_by is being set on object creation, but I don't see it being updated when objects are being updated. Is that because I'm missing something?

borrowd_items/models.py
chachan: that's for actual user deletion, not for user soft-deletes. Do you think this should be different?

chachan:
> Migrations pass on my local now!
> 
> I see that updated_by is being set on object creation, but I don't see it being updated when objects are being updated. Is that because I'm missing something?

I have reviewed all the places and Profile was missing a couple of places. Thanks, good catch.

ajk8 (approved):

PR #438, opened 2026-04-05
Add Docker support

joram:
Ike asked if borrowd PRs could work in VeilStream (I'm the dev of this tool).
For it to work, it needs to have a docker compose file in the borrowd repo.

With that, we can spin up preview environments based off any branch. here's a working copy of your stack running in a veilstream environment (TTL set for 7 days, by default it's shorter)
https://void-lion.env.veilstreamapp.com/

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,16 @@
+.git
+.gitignore
+.env
+node_modules/
+.venv/
+build/
+media/
+db.sqlite3
+staticfiles/
+__pycache__/
+*.pyc
+*.pyo
+.pytest_cache/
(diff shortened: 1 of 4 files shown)


ik-brwd:
We decided to avoid the added complexity for now - @joram will share more next time we talk!

joram:
> We decided to avoid the added complexity for now - @joram will share more next time we talk!

The avoidance of complexity is very sensible, those concerns passed through my mind as I wrote this PR.

Chat to you at one of the next Tenfolds :+1: @ik-brwd

Issue #41, opened 2025-05-10
Groups: Leave Group flow


rjendoubi:
Need a button for leaving Groups.

ik-brwd:
Naresh is taking a look at this...

PR #463, opened 2026-05-04
Nominate new moderator workflow

pushpitasaha:
## Summary

Implements moderator handoff when the last moderator leaves a group. The group is flagged, remaining members are notified, and the first member to accept becomes moderator.

## Linked Issue(s)
enhances #41 

## Type of Change
- [x] New feature

## Changes Made

- Added needs_moderator flag to BorrowdGroup
- Added signal to detect last moderator leaving (scoped to leave flow)
- Triggered notifications for remaining active members
- Implemented `BecomeModeratorView` with transaction locking
- Cleared notifications once a moderator is assigned
- Added banner in group detail view
- Added indicator in navigation + group list
- Added tests for moderator handoff flow

## How to Test

1. Create a group with at least 2 members (1 moderator, 1 member)
2. Log in as the moderator and leave the group
3. Log in as the remaining member
4. Verify:
   - Banner is shown on the group page
   - “!” indicator appears in navigation and group list
5. Click “Become moderator”
6. Verify:
   - User becomes moderator
   - Banner disappears
   - Indicators are cleared

## Media
nominate-mod-banner
nominate-mod-menu-dot
nominate-mod-menu-dot-group-list
nominate-mod-confirmation

## Checklist


borrowd_groups/models.py
```diff
@@ -93,6 +93,10 @@ class BorrowdGroup(Model):
         default=True,
         help_text="New members require Moderator approval to join the group",
     )
+    needs_moderator: BooleanField[Never, Never] = BooleanField(
+        default=False,
+        help_text="True when the group has no moderator and needs one.",
+    )
```
ajk8: Instead of implementing this as a boolean on the model, can it be a calculated property which checks to see if any of the users are moderators?

borrowd_groups/views.py
```diff
@@ -729,3 +735,55 @@ def post(
...
+    """
+    Allow first active member to claim moderator role.
+    """
+
+    def post(
+        self, request: HttpRequest, pk: int
+    ) -> HttpResponsePermanentRedirect | HttpResponseRedirect:
+        user: BorrowdUser = request.user  # type: ignore[assignment]
+
+        with transaction.atomic():
```
ajk8, line 783: Good call!

ajk8 (changes requested):

borrowd_groups/models.py
pushpitasaha: Updated - pushed the change and addressed this. Ready for another look!

ajk8 (approved):

PR #472, opened 2026-05-14
Fix/343/delete available items

### LINKED ISSUE #343: Cannot delete item
As a user, I should be able to delete an item that I own (whose status is currently "available"). 

In the below test case, you can see an item that should be able to be deleted - but it won't delete. 

I can delete items that are newly created (e.g. with no associated transaction record )  - is that the problem here, the fact that the item has been requested? (Does a transaction record block an item from being deleted?) 

https://github.com/user-attachments/assets/c00d02c9-5d9b-4a38-80f6-5524069559ae

Pejjp:
## Summary
Allows users to delete available items they own, even when those items have transaction history, by soft-deleting items and making active items the default query behavior.

## Linked Issue(s)
Closes #343 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Documentation

## Changes Made
- Updated item deletion to set `deleted_at` and `deleted_by` instead of hard-deleting the item row, preserving protected transaction history.
- Added an item manager/queryset so `Item.objects` returns active, non-deleted items by default.
- Added `Item.all_objects` for cases that need access to soft-deleted item records.
- Updated item views, filters, and inventory behavior to align with soft-delete semantics.
- Added regression tests for deleting available owned items with transaction history and preventing deletion of unavailable items.

## How to Test
1.  Manually create/request/cancel or reject an item so it has transaction history and is `AVAILABLE`.
2. Manually create/request/cancel or reject an item so it has transaction history and is `AVAILABLE`.
3. Confirm the owner can delete the item.
4. Confirm the deleted item no longer appears in inventory/search and its detail page is not accessible.

## Screenshots (if UI change)
N/A

## Checklist

## Risk / Notes for Reviewer
This changes the default `Item.objects` behavior to exclude soft-deleted items. Code paths that intentionally need deleted item records should use `Item.all_objects`. 
**Need confirmation on whether In-flight (non-available) items can be deleted too ?**

### DIFF
--- a/borrowd_items/filters.py
+++ b/borrowd_items/filters.py
@@ -65,12 +65,16 @@ class ItemFilter(FilterSet):  # type: ignore[misc]
         closer to a "public API".
         """
         if not hasattr(self, "_qs"):
-            qs: QuerySet[Item] = get_objects_for_user(
-                self.request.user,
-                ItemOLP.VIEW,
-                klass=Item,
-                with_superuser=False,
-            ).exclude(owner=self.request.user)
+            qs: QuerySet[Item] = (
+                get_objects_for_user(
+                    self.request.user,
+                    ItemOLP.VIEW,
+                    klass=Item,
+                    with_superuser=False,
+                )
+                .filter(deleted_at__isnull=True)
+                .exclude(owner=self.request.user)
+            )
             if self.is_bound:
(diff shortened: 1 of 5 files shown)


leilabl (approved):
LGTM!

Issue #494, opened 2026-06-02
Feature 2: Real-time Front-End Notification Interface (Client Experience)


iridescent-anupam:
Parent EPIC: [[EPIC]: Notification System](https://github.com/borrowd/borrowd/issues/492)

This feature implements the front-end user experience, bell notification hub, dropdown panel, and full notification history view. It has a hard dependency on Feature 1.

#### User Story 5: Real-Time Notification Bell & Badge

As an authenticated user, I want to see a bell icon with an unread badge in the nav bar that updates in real time while I’m on the page, so I know immediately when something needs my attention.

- **AC 5.1:** A bell icon is visible in the top-right nav bar for all authenticated users.
- **AC 5.2:** A dot nudge badge on the bell shows there are unread in-app notifications.
- **AC 5.3:** The badge updates in real time while the user is on the page (via WebSocket or SSE) — no page reload required.
- **AC 5.4:** If the real-time connection drops (network loss), the bell degrades gracefully: the badge shows the last-known count, and client-side JS automatically falls back to 30-second polling.

#### User Story 6: Notification Dropdown Panel

As a user, I want to click the bell to see a dropdown of my recent notifications and navigate directly to the relevant item, request, or group.

- **AC 6.1:** Clicking the bell opens a dropdown showing the most recent 10 notifications.
- **AC 6.2:** Each notification shows: a short message, relative timestamp (e.g. '2 hours ago'), and visual distinction between read and unread states.
- **AC 6.3:** Clicking a notification navigates to the relevant content and marks it as read.
- **AC 6.4:** A 'Mark all as read' action is available in the dropdown header.
- **AC 6.5:** A 'View all' link at the bottom opens the full `/notifications/` page.
- **AC 6.6:** Differentiate "Action-Required" notifications (mandatory types: `ITEM_REQUESTED`, `COLLECTION_ASSERTED`, `RETURN_ASSERTED`, `MEMBERSHIP_PENDING`) from standard "Informational" notifications using a distinct visual treatment (e.g., colored accent dot or bold styling).
- **AC 6.7:** If the target object (item, group, or transaction) of a notification is soft-deleted, preserve the notification text in the dropdown, but disable/grey-out the link click-through rather than showing a 404.

#### User Story 7: Notification History Page

As a user, I want a full notification history page where I can see all my past notifications, so I can review anything I missed.

- **AC 7.1:** `/notifications/` lists all notifications for the authenticated user, paginated (25 per page), newest first.
[...]

Issue #465, opened 2026-05-11
Feature: Ownership Transfer — Item Giveaway Flow


iridescent-anupam:
PRD Link: [Feature: #465 - Ownership Transfer — Item Giveaway Flow](https://docs.google.com/document/d/1oVvE6BdLzv3YNMvqX-bx_yfjiHO5K5ecrwr0zrO0tzg/edit?usp=sharing)
Related: [#437 Return of Item Flow](https://github.com/borrowd/borrowd/issues/437)
Design Reference: 

### Concept Note

**Why are we doing this?** Borrow'd is built on the idea that communities share — but sharing doesn't always end in a return. Sometimes a lender genuinely wants to give an item away: they're decluttering, they've moved cities, or they simply feel the borrower would get more use out of it than they would. Today, the platform has no structured way to handle this intent. A lender who wants to gift an item must awkwardly navigate a flow built entirely around the expectation of return — or worse, raise a sham dispute and resolve it with "Item not returned" just to remove it from their inventory. That is a trust-damaging hack.

**Who is going to use this?** Lenders who want to permanently transfer ownership of an item to a current borrower, and borrowers who need to formally accept (or decline) the gift and have it reflect correctly in their inventory.

**How will it be used?** This feature introduces two entry points: 
(1) During an active lending cycle — from the COLLECTED state, a lender can choose "Give away" instead of "Request item return". 
(2) From a dedicated “Giveaway Listing” — When adding an item, a lender can choose to mark it specifically as a giveaway, open to the community, where any borrower can claim it. In both cases, once the borrower accepts, ownership transfers permanently: the item is removed from the lender's inventory, added to the borrower's inventory as their own item, and both parties receive confirmation.

---

### Problem Statement

Currently, there is no legitimate, first-class path for a lender to voluntarily and permanently transfer ownership to a borrower. This creates three real friction points:

1. Lenders who want to give an item away are forced to either leave the transaction in COLLECTED indefinitely (cluttering their inventory and triggering false return-request reminders) or manipulate the dispute flow to remove it, which harms their own trust score optics.

2. Borrowers have no mechanism to formally accept ownership of a gifted item, meaning it appears in their inventory as something they still "owe back" rather than something they own.

3. Borrow'd misses a powerful community-building moment: generosity. Platforms like Freecycle and Buy Nothing have built loyal communities around the act of giving. Borrow'd currently has no hook into that behavior, leaving this segment of users underserved and the platform's network effects weaker.

---

### User Personas

[...]

PR #503, opened 2026-06-12
Feat 493 notifications backend

### LINKED ISSUE #493: Feature 1: Pluggable Routing Engine, Unified Preferences & Triggers (Backend & Preferences)
Parent EPIC: [[EPIC]: Notification System](https://github.com/borrowd/borrowd/issues/492)

This feature establishes the backend preferences, signal triggers, email integration, and pluggable architecture. It has no dependencies and is scheduled for V1 development.

#### User Story 1: Unified Notification Preferences

As a user, I want to control my notification settings easily, allowing me to customize channels (in-app vs email) grouped by logical categories, or turn off entire channels using master switches.

- **AC 1.1:** A `/settings/notifications/` page is accessible from account settings.
- **AC 1.2:** Settings are organized into logical categories in the UI:
    - Lending Lifecycle: `ITEM_REQUESTED` (Mandatory), `REQUEST_ACCEPTED`, `REQUEST_REJECTED`, `COLLECTION_ASSERTED` (Mandatory), `COLLECTION_CONFIRMED`, `RETURN_ASSERTED` (Mandatory), `RETURN_CONFIRMED`.
    - Group & Membership: `MEMBERSHIP_PENDING` (Mandatory), `MEMBERSHIP_APPROVED`.
    - Wishlist & Community: `COMMUNITY_REQUEST_POSTED`, `COMMUNITY_REQUEST_FULFILLED`.
    - Item Availability: `ITEM_AVAILABLE`.
- **AC 1.3:** Each optional notification type has two independent toggle switches: In-App and Email.
- **AC 1.4:** Surface two prominent master toggles at the top of the settings page:
    - Master In-App Toggle: Toggles all optional in-app notifications in bulk.
    - Master Email Toggle: Toggles all optional email notifications in bulk.
- **AC 1.5:** Each category section header features bulk in-app and email toggles to customize all optional items inside that specific group.
- **AC 1.6:** Critical transaction and moderation notifications (`ITEM_REQUESTED`, `COLLECTION_ASSERTED`, `RETURN_ASSERTED`, and `MEMBERSHIP_PENDING`) are strictly excluded from bulk overrides (both master switches and category switches) and individual switches. They are locked as always enabled, and their toggles are disabled in the UI with a label: "Required for transaction handshakes".
- **AC 1.7:** Store settings in `NotificationPreference(user, notification_type, in_app_enabled, email_enabled)`.
- **AC 1.8:** A database migration will initialize preference records for all existing users with both channels enabled by default.

#### User Story 2: Pluggable NotificationRouter & Delivery

As a developer, I want a pluggable router abstraction to dispatch notifications, so that we can easily support mobile push notifications in V2.

- **AC 2.1:** Implement a central `NotificationRouter` class that intercepts all notification events.
- **AC 2.2:** Register three pluggable backends: `InAppBackend` (V1 active), `EmailBackend` (V1 active), and `PushBackend` (V1 stub/no-op, to be activated in V2).
- **AC 2.3:** The router must check the `NotificationPreference` record before fanning out to each backend.
- **AC 2.4:** Wrap backend delivery in try/except blocks and log failures to Sentry. A failure in notification delivery must never abort the primary user transaction or database action.
- **AC 2.5:** A user must NEVER receive a notification triggered by their own actions. Filter out dispatches where `recipient == actor`.
- **AC 2.6:** Do not create duplicate notifications (same actor + recipient + verb + target) within a 10-minute window.
- **AC 2.7:** If a user receives >10 email notifications within an hour, suppress further individual emails and send a single summary digest instead. In-app notifications remain unthrottled.

#### User Story 3: Notification Triggers Across the Lending Lifecycle

As a lender or borrower, I want to be notified in real time when something relevant happens in a transaction or membership, so I can take timely action.

- **AC 3.1:** Generate notifications for the following events:
    - `ITEM_REQUESTED` -> Lender: "Maya requested your Drill"
    - `REQUEST_ACCEPTED` -> Borrower: "Your request for Drill was accepted"
    - `REQUEST_REJECTED` -> Borrower: "Your request for Drill was declined"
    - `COLLECTION_ASSERTED` -> Lender: "Maya says they have collected Drill. Please confirm."
    - `COLLECTION_CONFIRMED` -> Borrower: "Collection of Drill confirmed"
    - `RETURN_ASSERTED` -> Lender: "Maya says they have returned Drill. Please confirm."
    - `RETURN_CONFIRMED` -> Borrower: "Drill return confirmed. Thanks!"
    - `ITEM_AVAILABLE` -> Waitlisted user: "Drill is now available to borrow!"
    - `MEMBERSHIP_PENDING` -> Group moderator: "Maya wants to join [Group Name]. Review their request."
    - `MEMBERSHIP_APPROVED` -> New member: "You’ve been approved to join [Group Name]!"
- **AC 3.2:** All notification strings must follow the "Actor + Action + Object" format, using the actor's first name. Message strings must be warm, human, action-oriented, jargon-free, and limited to ≤80 characters. Timestamps must be relative (e.g. "3 minutes ago"), switching to absolute after 7 days.

#### User Story 4: Community Wishlist Notifications

As a group member, I want to be notified in real time when a group member posts a community request for an item I might have. As a requester, I want to be notified when a lender responds to my request by listing the item.

- **AC 4.1:** `COMMUNITY_REQUEST_POSTED` -> All shared-group members: "Maya is looking for Pressure Washer in [Group Name]"
- **AC 4.2:** `COMMUNITY_REQUEST_FULFILLED` -> Community requester: "Someone in [Group Name] listed Pressure Washer in response to your request!"
- **AC 4.3:** The `COMMUNITY_REQUEST_POSTED` broadcast is capped at groups with ≤50 active members for V1. For larger groups, delivery must be handled asynchronously via background queue tasks to protect request performance.

Pejjp:
## Summary

Implements the full in-app notification system for Borrow'd. This covers: 
- The `NotificationPreference` model (column-per-channel schema with `in_app_enabled, email_enabled, push_enabled`) 
- A `NotificationService` router that filters self-notifications, deduplicates within a 10-minute window, throttles email to 10/hour, and logs channel failures to Sentry. 
- Django signals are used to trigger each notification Type. 
- A reactive `/settings/notifications/` preferences page lets users toggle channels per notification type, with mandatory types locked. **This Page is accessible on this PR just via manual redirect to** `/settings/notifications/`
- A new temporary `/notifications/` inbox lists all in-app notifications with unread indicators, per-item dismiss, and mark-all-as-read.
- Real-time delivery: `htmx polling → SSE / WebSocket` migration path
  - The unread count badge and inbox updates currently rely on htmx polling (hx-trigger="every 30s" against a lightweight endpoint that returns the current unread count and swaps the badge fragment). 
  - This is intentional: polling requires zero new infrastructure, works correctly and is trivially testable .
  - When we move to **push-based delivery** (needed for messaging at some point), the migration surface is small.

## Linked Issue(s)

Closes #493

## Type of Change

- [ ] Bug fix
- [x] New feature
- [x] Refactor / cleanup
- [ ] Documentation

## Changes Made

| Area | Changes |
|---|---|
| `borrowd_notifications/models.py` | Added `NotificationType` enum with all lending, membership, and availability types; added `mandatory_types()` frozenset; added `NotificationPreference` model with per-channel boolean columns. |
| `borrowd_notifications/models.py` |`PushSubscription` — stores per-device push endpoint, `p256dh`, and `auth` key for each subscribed user. One row per browser/device; stale subscriptions (404/410 from the push service) are auto-deleted on send. |
| `borrowd_notifications/migrations/` | Added `0001_initial` to create the preferences table; added `0002_seed_default_preferences` to idempotently seed one row per existing user and notification type. |
| `borrowd_notifications/services.py` | Added `NotificationService.send_notification()` with self-notification filtering, deduplication, email throttling, per-channel backend dispatch, and Sentry capture on failure. |
| `borrowd_notifications/signals.py` | Added `post_save` handling for `Transaction` lending-lifecycle notifications; added `pre_save` and `post_save` handling for `Membership` pending, approved, and joined transitions. |
| `borrowd_notifications/views.py` | Added preferences context builder, toggle and bulk-toggle endpoints, inbox list, mark-read endpoints, and human-readable message formatting. |
| `templates/notifications/preferences.html` | Added Alpine.js single-component preferences UI with reactive master, category, and per-type toggles; state loads once from `json_script`; updates happen via `fetch` without page reload. |
| `templates/notifications/inbox.html` | Added paginated inbox with unread dot, relative timestamps via `naturaltime`, per-notification dismiss, and mark-all button. |
| `templates/includes/header.html` | Added `Notifications` sidebar link with bell icon. |
| `templates/notifications/messages/` | Added missing email templates. |
| `borrowd_notifications/tests.py` | Added preference-aware notification tests covering routing, channel errors, email throttling, and preference toggle endpoints. |
| Email throttle caps at 10 emails/hour per recipient and silently drops the remainder. | Once old emails age past the 1-hour mark : digest/summary with **NO RETRIES**   |

## How to Test

1. Run uv run manage.py migrate then uv run manage.py runserver
2. Log in as two users (user A and user B) who share a group
3. As user B, request to borrow an item from user A — user A should receive a notification in /notifications/
4. Accept the request as user A — user B receives a notification
5. Walk through collect/return steps and verify both parties receive the appropriate notifications
6. Visit /settings/notifications/ manually - toggle a channel off for a notification type; re-trigger that notification type and confirm it is no longer delivered via that channel
7. Toggle the category-level and master toggles and verify all children switch state instantly without a page reload
8. As a group moderator, approve a pending membership — the new member should receive "Membership approved" and existing members "Change to group membership"
9. Visit /notifications/, dismiss a notification, verify it loses its unread styling; use "Mark all as read" and verify the badge clears

## Screenshots (if UI change)

These are temporary UIs for /notifications/ and /settings/notifications/

Screenshot 2026-06-12 at 2 53 31 PM
Screenshot 2026-06-12 at 2 54 08 PM

### http://127.0.0.1:8000/settings/notifications/: 
Screenshot 2026-06-13 at 8 03 05 PM

https://github.com/user-attachments/assets/a6e27fa2-153c-4cc2-acec-1ae928687a60

https://github.com/user-attachments/assets/579d21a0-579b-48ee-ba9b-67c6ab7ac0d7

## Checklist

## Risk / Notes for Reviewer

- New users created after the seed migration won't have NotificationPreference rows; toggle_preference creates them on first toggle via get_or_create. The inbox and services handle the missing-row case gracefully (default = all channels off for optional types).
- Mandatory types (ITEM_REQUESTED, COLLECTION_ASSERTED, RETURN_ASSERTED, MEMBERSHIP_PENDING) bypass user preferences and always send via APP + EMAIL. Verify these cannot be toggled off in the UI (they render as disabled inputs).
- Community wishlist signals (COMMUNITY_REQUEST_POSTED, COMMUNITY_REQUEST_FULFILLED) are stubbed in the model and preferences UI **but have no signal wiring yet** — the CommunityRequest model doesn't exist. Notification types are present so preferences rows are seeded.


borrowd_notifications/channels.py
```diff
@@ -0,0 +1,154 @@
...
+                    context.get("respond_url")
+                    or context.get("item_url")
+                    or context.get("group_url")
+                    or f"{base_url}/notifications/"
+                ),
+            }
+        )
+
+        errors: list[str] = []
+        for sub in subscriptions:
```
Pejjp: Will be faster to use a Celery group() or chord() to fan out in parallel if the user has multiple connections.
If a group-wide event fires (e.g. a group announcement), we're pushing to N users × M devices

borrowd_notifications/services.py
```diff
@@ -1,156 +1,151 @@
...
-        text_message = render_to_string(f"notifications/{template_name}.txt", context)
-
-        send_mail(
-            subject=notification.description,
-            message=text_message,
-            html_message=html_message,
-            from_email=settings.DEFAULT_FROM_EMAIL,
-            recipient_list=[notification.recipient.email],
-            fail_silently=False,
+    def _is_duplicate(notification: Notification) -> bool:
```
Pejjp, line 95: Could be more precise on what is defined as duplicate.

borrowd_notifications/signals.py
```diff
@@ -1,202 +1,278 @@
...
+        case TransactionStatus.REJECTED:
+            notify.send(
+                instance.party1,
+                recipient=[instance.party2],
+                verb=NotificationType.ITEM_REQUEST_DENIED.value,
+                action_object=instance.item,
+                target=instance,
+                description="The item you requested is not available",
+            )
+        case TransactionStatus.COLLECTION_ASSERTED:
```
Pejjp, line 139: Transaction notifications now include `COLLECTION_ASSERTED`, `COLLECTION_CONFIRMED`, `RETURN_ASSERTED`, and `RETURN_CONFIRMED` states.
And proper notification triggers for collection and return hand-shake.
The counter party is notified and not the updater.

borrowd_notifications/signals.py
```diff
@@ -1,202 +1,282 @@
...
+                instance.updated_by,
+                recipient=[counterparty],
+                verb=NotificationType.RETURN_CONFIRMED.value,
+                action_object=instance.item,
+                target=instance,
+                description="{instance.item.name} return confirmed. Thanks for borrowing!",
+            )
+
+
+@receiver(pre_save, sender=Membership)
```
Pejjp, line 203: Membership status transitions are tracked with a pre_save hook so approval notifications only fire on `PENDING -> ACTIVE`

borrowd_notifications/signals.py
```diff
@@ -1,202 +1,282 @@
...
+                "status", flat=True
+            ).get(pk=instance.pk)
+        except Membership.DoesNotExist:
+            instance._previous_status = None  # type: ignore[attr-defined]
+    else:
+        instance._previous_status = None  # type: ignore[attr-defined]
+
+
+@receiver(post_save, sender=Membership)
+def send_membership_notifications(
```
Pejjp: Membership notifications now distinguish `pending` join requests, approvals from `pending` to active, and `active` member joined events.

borrowd_notifications/signals.py
```diff
@@ -1,202 +1,282 @@
...
+    if item.is_borrowable():
+        subscriptions = AvailabilitySubscription.get_active_subscriptions_for_item(item)
+        for subscription in subscriptions:
+            notify.send(
+                item.owner,
+                recipient=[subscription.user],
+                verb=NotificationType.ITEM_NOTIFY_WHEN_AVAILABLE.value,
+                action_object=item,
+                target=subscription,
+                description=f"{item.name} is now available",
```
Pejjp, line 54: `description` fields are stored but not used by the channels, will be removed or replaced with: `NotificationType.ITEM_NOTIFY_WHEN_AVAILABLE.message_template`

vguzman812:
UX issues:

1. Logging in to an already created account shows that I have unread notifications, but none show up.
Screenshot 2026-06-16 at 5 09 11 AM

2. There is no way to get to settings/notifications without manually typing it in that I can tell of. 

3. Going through onboarding and enabling everything, the notification preference for "a member joined a group you're a part of" is not affected by the buttons to tall all notifcations of a category.
Screenshot 2026-06-16 at 5 15 28 AM

4. There's no way to continue past the notifications settings page in onboarding without the user clicking a different link from the hamburger menu. This means a new user can't enable any notifications besides the base ones.
Screenshot 2026-06-16 at 5 18 04 AM

5. There's no save button or similar for notification preferences, and refreshing the page restores all preferences to their original state (none but the required ones selected)

6. There's no bell or notifications badge anywhere indication there is a notification for the user, but this is one of the requirements of #494 (user story 5), which I assume this PR is meant to tackle in addition to #493
Screenshot 2026-06-16 at 5 21 17 AM

7. Clicking dismiss does not remove the notification, though it does mark it as read. As a basic user, I would assume this would delete the notification, not set it as read.
Screenshot 2026-06-16 at 5 23 07 AM

8. When a notification is disabled, I still receive the new notification badge even if the notification does not show up.
Screenshot 2026-06-16 at 5 28 08 AM

9. Notifications require a refresh to show up.
https://github.com/user-attachments/assets/8bf4f316-2384-4d5c-9de0-19198ca0e370

Pejjp:
> 5\. but

Notification Issues — Updated Diagnostic


| #   | Issue | Status |
|---|---|---|
| 1 & 8 | Unread count mismatch / badge for disabled types | FIXED                                       
| 2 | No link to notification settings | NOT FIXED This will be part of #494 and will be implemented once the designs are available.                                  
| 3a | Stale seed data (GROUP_MEMBER_JOINED) | FIXED                         
| 3b | Bulk toggle loop | FIXED                       
| 4 | No continue button in onboarding | NOT FIXED  This can be added to #494 as well
| 5 | init_new_user_preferences broken | FIXED
| 6 | No bell/badge in navbar | NOT FIXED       This can be added to #494 as well
| 7 |"Delete and Delete all notifications "  | ADDED 
| 9 | Requires page refresh | PARTIAL — inbox polls. Wee need a real-time push channel.

vguzman812 (changes requested):
This is quite a large PR. Before I went into it manually, I had Claude do a general review as well as a review using the acceptance criteria in #493. There are quite a few things that were reported. Not all of these might necessarily need to be addressed, but there are some more critical ones.
At the very least, all #493 acceptance criteria need to be met where possible, though not everything can be met at the moment. For instance, the community wishlist requirements can't be accomplished at the moment since that feature is not implemented yet.
Please see the report below for full details.

---

## Logic/correctness

### 1. Email throttle is effectively dead - `emailed` is never persisted
- `EmailNotificationStrategy.send` sets `payload.notification.emailed = True` in memory (`channels.py:60`), but `NotificationService.send_notification` only writes the `data` column back: `Notification.objects.filter(pk=...).update(data=data)` (`services.py:163`). `emailed` is never saved. Per the [Django docs on `QuerySet.update()`](https://docs.djangoproject.com/en/5.2/topics/db/queries/#updating-multiple-objects-at-once), it "performs an SQL update query for the specified fields" only and "does not run any `save()`  methods on your models, or emit the `pre_save` or `post_save` signals" - so passing data=...` updates the `data` column and nothing else; the in-memory `emailed=True` is discarded.
- `_is_email_throttled` counts `Notification.objects.filter(..., emailed=True, ...)` (`services.py:101-109`). Since `emailed` stays at its default of `False` in the DB, that count is always 0, so the hourly cap (AC 2.7) never fires in production.
- Every throttle test (`tests.py:889-971`) mocks `_is_email_throttled` to return `True`, so the broken predicate is never exercised. CI passes while the feature is non-functional.
- Fix: persist `emailed` (include it in the `update_fields`/`.update()` call after the email backend runs), and add at least one test that drives `_is_email_throttled` through real rows instead of mocking it. Note if you switch to `instance.save(update_fields=[...])`: Django [saves only the named fields](https://docs.djangoproject.com/en/5.2/ref/models/instances/#specifying-which-fields-to-save), so `emailed` has to be in that list explicitly.

### 2. New-user preference defaults contradict AC 1.8 and the seed migration
- `NotificationPreference.init_new_user_preferences` (`models.py:283-300`) sets optional types to `in_app_enabled=mandatory | False` / `email_enabled=mandatory | False`. For optional types `mandatory` is `False`, so both channels are created disabled. New users are opt-in (everything off except mandatory).
- AC 1.8 says preferences should initialize "with both channels enabled by default," and migration `0002_seed_default_preferences` does exactly that for existing users (`in_app_enabled=True, email_enabled=True`).
- Net result: existing users are opt-out (all on), new users are opt-in (all off). Two different defaults for the same product. `tests.py:1260` codifies the opt-in behavior, so this looks deliberate, but it conflicts with both the PRD and the migration. The team should pick one and make them agree.
- Side note: `mandatory | False` is a confusing bitwise-OR on a bool; it is just `mandatory`. Replace with a plain boolean for readability.

### 3. Community wishlist triggers (User Story 4) are not implemented
- `COMMUNITY_REQUEST_POSTED` and `COMMUNITY_REQUEST_FULFILLED` exist as `NotificationType` values and appear in the settings UI, but nothing ever fires them - no signal or service call references them (`grep` finds zero producers).
- AC 4.1 / 4.2 / 4.3 (including the <=50-member broadcast cap and async fan-out for larger groups) are unfulfilled.

### 4. Summary digest is scheduled but never sent
- When email is throttled, `_schedule_summary_digest` writes a `summary_digest` dict into `notification.data` (`services.py:111-117, 144-146, 160-161`). Nothing ever reads that back and sends a digest email - there is no task, cron, or worker.
- AC 2.7 requires "send a single summary digest instead." Today the extra emails are just dropped. This is a stub; flag it as such or wire up the send.

### 5. `counter_party()` is computed unconditionally in the transaction signal
- `send_transaction_notifications` calls `counterparty = instance.counter_party(instance.updated_by)` at the top, before the `match` (`signals.py:91`). `counter_party` raises `ValueError` if `updated_by` is not party1 or party2 (`borrowd_items/models.py:1029-1036`).
- This runs on *every* `Transaction.save()`, including statuses that send nothing (e.g. `RESOLVED` from `force_resolve`, which sets `updated_by=resolved_by`, `models.py:1071`). Today resolvers happen to be a party, so it works - but any save where `updated_by` is the system user or a non-party moderator would raise and roll back the whole transaction save. That is a landmine.
- Fix: move the `counter_party(...)` call into the specific case branches that need it (COLLECTION_ASSERTED / COLLECTED / RETURN_ASSERTED / RETURNED), which already only fire for party-driven transitions.

### 6. `RETURN_CONFIRMED` description is a missing f-string
- `signals.py:154`: `description="{instance.item.name} return confirmed. Thanks for borrowing!"`. No `f` prefix, so the literal braces ship through. This string is used as the email subject (`channels.py:56` passes `notification.description` as `subject`), so the borrower gets a subject line reading `{instance.item.name} return confirmed.`
- Every sibling case uses an f-string; this one was missed.

### 7. `delete_app_notification` sets a method, not a field
- `views.py:353`: `notification.delete = True`. There is no `delete` field on the django-notifications `Notification` model - `delete` is the instance method, so this just shadows it on a throwaway instance and is never persisted (it is not in `update_fields=["data", "unread"]`, and Django only writes the [fields named in `update_fields`](https://docs.djangoproject.com/en/5.2/ref/models/instances/#specifying-which-fields-to-save)). The actual "delete" works by popping `APP` out of `data["channels"]`. The line is dead and misleading; remove it. If a real soft delete was intended, the library field is `deleted` (and needs `NOTIFICATIONS_SOFT_DELETE`).

### 8. Inbox 30s poll swaps a full page into a fragment
- `inbox.html:43-46`: the card div has `hx-get="{% url 'notification-inbox' %}"`, `hx-trigger="every 30s"`, `hx-swap="outerHTML"`. The endpoint returns the whole page (it `extends layouts/default.html`), so every 30s htmx replaces the card with an entire document - nested `<html>`/header inside the card. With [`hx-swap="outerHTML"`](https://htmx.org/attributes/hx-swap/) htmx "replaces the entire target element with the response," so the whole returned document lands in the card; either return a partial template, or add [`hx-select`](https://htmx.org/attributes/hx-select/) to "select the content you want swapped from a response" (a CSS selector pulling just the list).

### 9. Click-through and soft-deleted-target handling are missing
- `_notification_target_url` is computed per row in the view (`views.py:296-300, 314`) but `inbox.html` never renders it - notifications are not links, so AC 6.3 ("clicking navigates and marks read") is unmet and the computed URL is dead.
- AC 6.7 / 7.4 (preserve text but disable the deep link when the target is soft-deleted) is not implemented at all.

### 10. Transaction notifications fire on every save, not on a status transition
- `send_transaction_notifications` (`signals.py:85-92`) matches on `instance.status` for every `post_save`, with no `created` check and no previous-status capture. Any later save of a Transaction in the same status re-enters the matching branch and emits the same notification again.
- The only thing keeping that from duplicating is the 10-minute dedup window (`services.py:84-98`); a second save of an unchanged status >10 minutes later re-sends the email/in-app notification.
- The membership signal already does this correctly via `capture_membership_previous_status` (`signals.py:184-197`). Mirror that pattern: capture the prior status in `pre_save` and only notify on an actual transition.

### 11. Collection/return copy assumes the borrower asserted
- The state machine lets either party assert return - in the `COLLECTED` branch the owner also gets `MARK_RETURNED` (`models.py:487-489`). But the context builder for `COLLECTION_ASSERTED` / `RETURN_ASSERTED` hardcodes `requester_name = transaction.party2...` (`models.py:168-179`), and the templates greet the owner and say the requester performed the action.
- So if the owner asserts return, the borrower (party2) is told `"{party2-name} says they have returned ..."` - naming the recipient as the actor. Derive the copy from `notification.actor` / `notification.recipient` instead of fixed party1/party2 roles.

---

## Security

Mostly solid. Notable points:

### Risks
- Silent failures never reach Sentry (AC 2.4). Each strategy wraps its own body in `try/except` and records the error into `data` (`channels.py:44-64, 82-88, 92-101`). Because the strategy swallows the exception and returns normally, the service-level `sentry_sdk.capture_exception` (`services.py:155-157`) never runs for an email/app/push send failure. So a broken SMTP server is invisible in Sentry - the only trace is a JSON blob on the notification row. AC 2.4 explicitly wants these logged to Sentry.
- Postgres `data::jsonb` cast is fragile (`views.py:264-271`). `data` is stored as text. NULL casts fine, and current writers only ever store valid JSON, but a single row with non-JSON text in `data` would make the whole inbox query raise for that user. Worth a defensive guard (e.g. validate/normalize on write, or filter to rows that look like JSON) given the column is shared with django-notifications-hq.
- "Member joined" broadcasts to non-active members (`signals.py:227,249`). Recipients are `instance.group.users.exclude(id=instance.user.pk)`, and `BorrowdGroup.users` (`borrowd_groups/models.py:96`) is the unfiltered M2M through Membership, which spans PENDING / SUSPENDED / BANNED / ENDED. So pending/suspended/banned users get told someone joined a group they aren't active in - a privacy leak and inconsistent with the rest of the code. Scope these to `MembershipStatus.ACTIVE`, the way the MEMBERSHIP_PENDING moderator query already does a few lines up (`signals.py:207-214`).
- PUSH records success without delivering (`channels.py:91-101`, `services.py:80`). The preferences UI exposes push toggles and the router adds PUSH to the channel set when `push_enabled`, but `PUSHNotificationStrategy.send` marks the channel SUCCESS unconditionally without sending anything. A user can "enable push" and the audit data on the row reports a success that never happened. Until push ships, hide the push column or record it as PENDING/unsupported rather than SUCCESS.

---

## Maintainability / style

- `NotificationType._get_template_context_for` (`models.py:126-246`) is a long match/if-chain carrying a `# TODO simplify` of its own. It reaches deep into Transaction/Group/Subscription internals. The `ITEM_SUBSCRIPTION` and `ITEM_NOTIFY_WHEN_AVAILABLE` branches (`models.py:228-245`) are byte-for-byte identical and can collapse into one.
- `instance.public = False` in the `Notification` post_save handler (`signals.py:77`) is set then never saved - it has no effect. Either persist it or drop it.
- `print()` left in production paths: `PUSHNotificationStrategy._send_push` (`channels.py:104`) and the `decode` error branch (`borrowd/util.py`). Use logging.
- Message copy vs AC 3.2 (<=80 chars, "Actor + Action + Object", first name):
  - `ITEM_RETURN_REQUESTED` template (`models.py:62-65`) is ~107 chars.
  - `ITEM_DISPUTED` = "This item has escalated to a dispute!" - no actor, no object name.
- `mandatory_types()` (`models.py:106-116`) includes `ITEM_RETURN_REQUESTED`, which is not in the PRD's mandatory set (AC 1.6 lists only ITEM_REQUESTED, COLLECTION_ASSERTED, RETURN_ASSERTED, MEMBERSHIP_PENDING). Intentional? If so, note it; if not, it locks a type the PRD wanted optional.
- Migration churn: `0002` seeds a hardcoded type list, `0003` renames/deletes some of them, `0004` re-alters the choices. The `0002` hardcoded `ALL_NOTIFICATION_TYPES` duplicates the model and went stale within the same PR. Fine to leave (migrations are historical), just a sign the type set was not settled before seeding.
- Dispatch is fully synchronous on the request thread: each recipient is its own Notification -> `post_save` -> `on_commit` -> `send_notification`, which runs preference + dedup queries and a blocking `send_mail` per recipient. Group broadcasts already fan out to every member, so a large group turns one join into O(members) blocking work inside the request. It's recoverable because everything funnels through `NotificationService.send_notification` - make that enqueue-and-return before group-broadcast/messaging scale, and avoid building synchronous assumptions on top of it.
- `notification_type` / `Notification.verb` stores an English phrase ("Item requested", `models.py:27`). That same string is the preference key and is fed to `NotificationType(verb)` on historical rows, so renaming any copy orphans every preference row and breaks lookups on old notifications. Prefer a stable code ("ITEM_REQUESTED") with the human label kept separate.
- `Transaction.counter_party` raises a bare `ValueError` (`borrowd_items/models.py:1035`). Per AGENTS.md, domain failures should subclass `BorrowdException` (e.g. in `borrowd_items/exceptions.py`).
- `toggle_preference` create-defaults are `{in_app:True, email:True}` (`views.py:209-213`), but the page renders a rowless optional type as all-off. Flipping only PUSH on for such a type silently also enables in-app and email. Creation defaults should mirror the displayed defaults.
- `_is_duplicate` (`services.py:84-98`) excludes only the row's own pk, so two identical notifications created in the same transaction can each see the other and both get dropped, leaving the user with nothing. Tighten the key or the exclusion.
- `NotificationData.icon` is always `None` and the `status` property is never read (`models.py`). Dead surface - use it or drop it.
- Typos throughout (cosmetic but plentiful): "witch" (`services.py:47`), "mendatory" (`views.py:138`), "missmatch"/"goup"/"strategie" in commit messages, "counterpartie" test name (`tests.py:601`).
- `NOTIFICATIONS_ENABLED` global kill switch from the PRD "Fall Back" section is not implemented. Optional for V1, but it is a named requirement.

---

## Tests

- Good breadth: 59 tests covering self-notification suppression, moderator targeting, per-channel preference dispatch, mandatory bypass, failure recording, toggle/bulk-toggle endpoints, inbox app-channel filtering, and delete behavior. Plus `tests/test_return_flows.py` for the dispute/return state machine.
- Gap: the throttle/digest tests (`tests.py:889-971`) all mock `_is_email_throttled`, so the real counting logic - which is broken (finding # 1) - is never tested. Add a test that creates >10 emailed notifications and asserts the 11th is suppressed, driving the predicate for real.
- Gap: no test asserts that an email backend success actually persists `emailed=True` on the row (would have caught # 1).
- Gap: `counter_party` is only exercised via party-driven flows; no test saves a Transaction with a non-party `updated_by` (would surface # 5).

---

## PRD acceptance-criteria coverage (Feature 493)

### User Story 1 - Unified Notification Preferences

- 🟡 **AC 1.1 - Partial.**
  - Full AC: "A `/settings/notifications/` page is accessible from account settings."
  - The page exists (`notification_preferences_view` at `/settings/notifications/`), but nothing links to it from account settings - the only nav link added points to the inbox (`header.html`), so it's reachable only by typing the URL.
- 🟢 **AC 1.2 - Met.**
  - Four categories present in `NOTIFICATION_CATEGORIES`.
- 🟢 **AC 1.3 - Met.**
  - Each optional type has independent In-App and Email toggles.
- 🟢 **AC 1.4 - Met.**
  - Master In-App and Master Email toggles present.
- 🟢 **AC 1.5 - Met.**
  - Category-level bulk in-app/email toggles present.
- 🟡 **AC 1.6 - Met (caveat).**
  - Mandatory types locked for APP/EMAIL (403 server-side + disabled in UI). Caveat: PUSH is still toggleable/bulk-applied for mandatory types.
- 🟢 **AC 1.7 - Met.**
  - Stored in `NotificationPreference(user, notification_type, in_app_enabled, email_enabled)` (plus `push_enabled`).
- 🟡 **AC 1.8 - Partial.**
  - Full AC: "A database migration will initialize preference records for all existing users with both channels enabled by default."
  - Migration 0002 seeds existing users all-on, but: (a) new users get optional types created disabled, so the default diverges by cohort (finding # 2); (b) the newer types `Item disputed` / `Item return requested` were never backfilled, so existing users have no row for those.

### User Story 2 - Pluggable NotificationRouter & Delivery

- 🟢 **AC 2.1 - Met.**
  - Central router present (named `NotificationService`, not `NotificationRouter`).
- 🟢 **AC 2.2 - Met.**
  - Three backends registered: App + Email active, Push as a no-op stub.
- 🟢 **AC 2.3 - Met.**
  - Router checks `NotificationPreference` before fan-out (`_get_enabled_channels`).
- 🟡 **AC 2.4 - Partial.**
  - Full AC: "Wrap backend delivery in try/except blocks and log failures to Sentry. A failure in notification delivery must never abort the primary user transaction or database action."
  - The never-abort half holds (dispatch runs in `transaction.on_commit`), but the Sentry half doesn't: each strategy swallows its own exception into `data`, so the service-level `sentry_sdk.capture_exception` never fires for a real send failure (see Security/Risks).
- 🟢 **AC 2.5 - Met.**
  - `actor == recipient` guard plus recipient exclusion (`services.py:121`).
- 🟢 **AC 2.6 - Met.**
  - 10-minute dedup on actor + recipient + verb + target (`services.py:84-98`).
- 🔴 **AC 2.7 - Not functional.**
  - Full AC: "If a user receives >10 email notifications within an hour, suppress further individual emails and send a single summary digest instead. In-app notifications remain unthrottled."
  - The cap never engages because `emailed` is never persisted (finding # 1), and the digest is scheduled into `data` but never sent (finding # 4).

### User Story 3 - Notification Triggers Across the Lending Lifecycle

- 🟡 **AC 3.1 - Met (caveat).**
  - Triggers wired for transaction + membership + availability lifecycle events. Caveat: the `RETURN_CONFIRMED` copy ships a broken subject (finding # 6).
- 🟡 **AC 3.2 - Partial.**
  - Full AC: "All notification strings must follow the 'Actor + Action + Object' format, using the actor's first name. Message strings must be warm, human, action-oriented, jargon-free, and limited to <=80 characters. Timestamps must be relative (e.g. '3 minutes ago'), switching to absolute after 7 days."
  - Gaps: some templates exceed 80 chars (`ITEM_RETURN_REQUESTED` ~107), `ITEM_DISPUTED` doesn't follow the actor+action+object format, and `naturaltime` stays relative forever (no switch to absolute after 7 days).

### User Story 4 - Community Wishlist Notifications

- 🔴 **AC 4.1 - Not met.**
  - Full AC: "`COMMUNITY_REQUEST_POSTED` -> All shared-group members: 'Maya is looking for Pressure Washer in [Group Name]'."
  - The type exists but nothing fires it (finding # 3).
- 🔴 **AC 4.2 - Not met.**
  - Full AC: "`COMMUNITY_REQUEST_FULFILLED` -> Community requester: 'Someone in [Group Name] listed Pressure Washer in response to your request!'."
  - Type exists, no producer.
- 🔴 **AC 4.3 - Not met.**
  - Full AC: "The `COMMUNITY_REQUEST_POSTED` broadcast is capped at groups with <=50 active members for V1. For larger groups, delivery must be handled asynchronously via background queue tasks to protect request performance."
  - No cap and no async path exist.

Pejjp:
Thanks for the detailed comments.
Here are the fixes I propose for this PR to keep it light.
I will create new PRs for the other requested changes.

### Changes made
| # | What | Files |
|---|------|-------|
| 1 | `emailed` persisted after successful send + 2 real throttle tests | `services.py`, `tests.py` |
| 5+6 | `counter_party()` moved into branches that need it; `RETURN_CONFIRMED` f-string fixed | `signals.py` |
| 10 | Transaction notifications only fire on status transition (pre_save captures previous status) | `signals.py` |
| 7 | Dead `notification.delete = True` line removed | `views.py` |
| 8 | htmx 30s poll uses `hx-select` to avoid swapping full page into fragment | `inbox.html` |
| member joined | `GROUP_MEMBER_JOINED` scoped to active members only | `signals.py` |
| PUSH | Strategy is a true no-op; push column hidden from preferences UI | `channels.py`, `preferences.html` |
| public | `public=False` persisted via `.update()` on notification creation | `signals.py` |
| defaults | `toggle_preference` create-defaults match the all-off UI state | `views.py` |
| AC 2.4 | Strategy exceptions re-raised so Sentry captures them | `channels.py` |
| AC 1.1 | Notification settings linked from profile/account settings page | `profile.html` |
| 11 | Collection/return notification copy hardcoder asserts return, borrower is told they themselvesreturned the item |`models.py` (`_get_template_context_for`) |
| AC 1.8 | New-user preference defaults (all-on) like existing-user  | `models.py`, migration |
| AC 3.2 | Relative timestamps  switch to absolute after 7 days | templates |
| AC 2.7 | Summary digest is stubbed into `data` actually sent — added a cron | `services.py`, new worker |

## For other PRs

| # | What | Notes | Files |
|---|------|-------|-------|
| 9 | Inbox notifications are not clickable — `target_url` computed but never rendered because it was intended for redirect on push-notifications ; no mark-as-read on click (AC 6.3) | Near-term | `inbox.html`, `views.py` |
| AC 6.7/7.4 | Preserve notification text but disable deep link when target is soft-deleted | Near-term | `views.py`, `inbox.html` |
| AC 4.x | `COMMUNITY_REQUEST_POSTED` / `COMMUNITY_REQUEST_FULFILLED` have no producers — blocked on community wishlist feature | Blocked on feature #465  | `signals.py` |
| — | `NOTIFICATIONS_ENABLED` global kill switch from PRD "Fall Back" section | Optional V1 requirement | settings, `services.py` |
| — | Async fan-out for large group broadcasts — current dispatch is synchronous on the request thread | Separate infrastructure | `signals.py`, new worker |

vguzman812:
Thanks for the fixes! As for the deferred aspects you brought up, I think we can (and should) make some of those fixes now before we push this pr into staging. 

For AC 1.8: If we're seeding existing users as opt-out (all notifications enabled by default), then a logical jump would be that all users should have all notifications enabled by default. So let's go ahead with that assumption for now.

For AC 2.7: We can make cron jobs by editing `.platform.app.yaml`. see [here](https://support.platform.sh/hc/en-us/articles/11975558123666-How-to-configure-a-cron-job).

For AC 3.2: I can see the point, but if we're trying to solve #493, this should be done in this PR. 

You have a lot of migrations, but this is the initial introduction of this feature, so really only 2 migrations are needed at this point: making the model and backfilling the users. The changes introduced in migration 3-5 can all be squashed into these two. Additionally, this also allows you to fix the `NotificationType` codes without another migration.

borrowd_notifications/models.py
```diff
@@ -0,0 +1,401 @@
...
+            mandatory = (
+                True
+                if notification_type in NotificationType.mandatory_types()
+                else False
+            )
+
+            NotificationPreference.objects.create(
+                user=user,
+                notification_type=notification_type,
+                in_app_enabled=mandatory | False,
```
vguzman812: You'realready setting `mandatory` to True or False above, so the extra `| False` is unnecessary.

borrowd_notifications/models.py
```diff
@@ -0,0 +1,401 @@
...
+    in_app_enabled: models.BooleanField[bool, bool] = models.BooleanField(default=True)
+    email_enabled: models.BooleanField[bool, bool] = models.BooleanField(default=True)
+    push_enabled: models.BooleanField[bool, bool] = models.BooleanField(default=False)
+
+    @staticmethod
+    def init_new_user_preferences(user: BorrowdUser) -> None:
+        """When a new user is created, the preferences must be initialised."""
+
+        for notification_type in NotificationType.values:
+            mandatory = (
```
vguzman812: This can be simplified to `mandatory = notification_type in NotificationType.mandatory_types()`

borrowd_notifications/models.py
```diff
@@ -0,0 +1,401 @@
...
+        "Item notify when available",
+        "{item_name} is now available to borrow",
+    )
+    ITEM_SUBSCRIPTION = (
+        "Item subscription",
+        "You have subscribed to be notified when {item.name} becomes available. We will let you know when it does!",
+    )
+    ITEM_RETURN_REQUESTED = (
+        "Item return requested",
+        "The owner of {item_name} has requested to get it back. Please coordinate with them the return of this item.",
```
vguzman812: This is ~95 characters not including `{item_name}`, which could be quite long. AC 3.2 in #493 says notifications shoul be "limited to ≤80 characters." Same with `ITEM_SUBSCRIPTION`.

borrowd_notifications/models.py
```diff
@@ -0,0 +1,401 @@
...
+    RETURN_CONFIRMED = "Return confirmed", "Return of {item_name} has been confirmed"
+
+    # Item availability
+    ITEM_NOTIFY_WHEN_AVAILABLE = (
+        "Item notify when available",
+        "{item_name} is now available to borrow",
+    )
+    ITEM_SUBSCRIPTION = (
+        "Item subscription",
+        "You have subscribed to be notified when {item.name} becomes available. We will let you know when it does!",
```
vguzman812: should this be `item.name` or `item_name`?

borrowd_notifications/models.py
```diff
@@ -0,0 +1,401 @@
...
+    )
+    ITEM_RETURN_REQUESTED = (
+        "Item return requested",
+        "The owner of {item_name} has requested to get it back. Please coordinate with them the return of this item.",
+    )
+    ITEM_DISPUTED = ("Item disputed", "This item has escalated to a dispute!")
```
vguzman812: AC 3.2 says "All notification strings must follow the "Actor + Action + Object" format, using the actor's first name"

borrowd_notifications/signals.py
```diff
@@ -118,14 +134,41 @@ def send_transaction_notifications(
...
                 description="The item you requested is not available",
             )
+        case TransactionStatus.COLLECTION_ASSERTED:
+            notify.send(
+                instance.updated_by,
+                recipient=[instance.counter_party(instance.updated_by)],
+                verb=NotificationType.COLLECTION_ASSERTED.value,
+                action_object=instance.item,
+                target=instance,
+                description=f"{instance.updated_by} says they've collected {instance.item.name}. Please confirm.",
```
vguzman812: `instance.updated_by` will return the user's username, which is the user's email address. Definitely a leak to fix. Use `instance.updated_by.first_name`.

borrowd_notifications/models.py
```diff
@@ -0,0 +1,401 @@
...
+    def __str__(self) -> str:
+        return self.name.lower()
+
+    @property
+    def message_template(self) -> str:
+        return str(self.label)
+
+    # TODO simplify the logic to make it cleaner.
+    @staticmethod
+    def _get_template_context_for(notification: notifications) -> Dict[str, Any]:
```
vguzman812: `notification: notifications` should be `notification: Notification`. Otherwise, the type is a module.

borrowd_notifications/services.py
```diff
@@ -1,176 +1,170 @@
...
+    ) -> set[ChannelType]:
+        if notification_type in NotificationType.mandatory_types():
+            return {ChannelType.APP, ChannelType.EMAIL}
+
+        try:
+            pref = NotificationPreference.objects.get(
+                user=user, notification_type=notification_type.value
+            )
+        except NotificationPreference.DoesNotExist:
+            return set()
```
vguzman812: This should do something other than drop silently. If someone adds a future optional notification type and doesn't backfill it properly (unlikely it would pass PR, but still), there'd be no indication that anything was wrong except for the given user. We should probably default a missing row to enabled rather than not.

borrowd_notifications/signals.py
Pejjp: the notifications sent use `template = NotificationType(notification.verb).message_template`, not the description fields.
Best to remove them ?

borrowd_notifications/services.py
```diff
@@ -1,176 +1,170 @@
...
+            channels.add(ChannelType.APP)
+        if pref.email_enabled:
+            channels.add(ChannelType.EMAIL)
+        if pref.push_enabled:
+            channels.add(ChannelType.PUSH)
+        return channels
+
+    @staticmethod
+    def _is_duplicate(notification: Notification) -> bool:
+        return bool(
```
vguzman812: .exists() already returns a bool, so wrapping in `bool()` is redundant.
```
(method) def exists() -> bool
Return True if the QuerySet would have any results, False otherwise.
```

borrowd_notifications/views.py
```diff
@@ -1,3 +1,377 @@
...
+    context: dict[str, Any] = {}
+    if isinstance(notification.data, dict):
+        context = notification.data.get("context", {})
+    try:
+        return template.format(**context)
+    except KeyError:
+        return str(notification.verb)
+
+
+def _notification_target_url(notification: Notification) -> str | None:
```
vguzman812: If we're not going to use this now (it's part of #494), then it should be removed.

borrowd_notifications/views.py
```diff
@@ -1,3 +1,377 @@
...
+    """Filter to notifications that were dispatched via the APP channel.
+
+    The `data` field from django-notifications-hq is stored as plain text (not
+    native jsonb), so we must branch on the DB vendor: SQLite supports
+    JSON_TYPE(); PostgreSQL requires an explicit ::jsonb cast plus the -> operator.
+    """
+    if connection.vendor == "postgresql":
+        return qs.annotate(
+            app_channel=RawSQL(
+                "(data::jsonb) -> 'channels' -> 'APP'",
```
vguzman812: one row with '' or any non-JSON text raises `DataError: invalid input syntax for type json` with Postgres when the user navigates to their inbox. Why not just do something like `return qs.filter(data__contains='"APP":')`?

What do you think about querying on a simple bool field instead of trying to parse a json blob. Something like `shown_in_app` so you could just do `qs.filter(shown_in_app=True)`

borrowd_notifications/services.py
```diff
@@ -1,176 +1,170 @@
...
-            elif notification.verb == NotificationType.ITEM_NOTIFY_WHEN_AVAILABLE.value:
-                context.update(
-                    {
-                        "subscriber_name": subscription.user.profile.full_name(),
-                        "item_name": subscription.item.name,
-                        "item_url": settings.BASE_URL
-                        + reverse("item-detail", args=[subscription.item.pk]),
-                    }
+    def _is_email_throttled(recipient: BorrowdUser) -> bool:
+        return bool(
```
vguzman812, line 117: >= will return a bool, so no need for bool()

borrowd_notifications/services.py
Pejjp: because of mypy I get error: Returning Any from function declared to return "bool"  [no-any-return]

borrowd_notifications/services.py
```diff
@@ -1,176 +1,170 @@
...
+            Notification.objects.filter(
+                actor_content_type=notification.actor_content_type,
+                actor_object_id=notification.actor_object_id,
+                recipient=notification.recipient,
+                verb=notification.verb,
+                target_content_type=notification.target_content_type,
+                target_object_id=notification.target_object_id,
+                timestamp__gte=timezone.now() - _DEDUP_WINDOW,
+            )
+            .exclude(pk=notification.pk)
```
vguzman812: This is a good exclusion generally, but I have a nitpick due to community requests coming eventually.
Imagine a scenario where we have a community wishlist, and `COMMUNITY_REQUEST_POSTED` sends a notification out to "all shared-group members" (AC 4.1). Now imagine there's a super active user who shares 3 groups with the requester. The query that gathers those members has to reach across all 3 groups, and that super-active person shows up in the list once per shared group. Also, assume whoever coded this does not account for uniqueness.
The super-active user gets added 3 times to this community notification event, which means 3 identical notifications, all created in the same transaction. Excluding by pk will mean all 3 notifications are excluded. So instead of one notification, the super-active member gets nothing. 

To avoid that, I think excluding based on the earliest notification timestamp, then on pk is a stronger approach here, though I'm open to suggestions.

borrowd_notifications/services.py
Pejjp: Added fallback to default = true

borrowd_notifications/services.py
Pejjp: raises a mypy error

vguzman812 (changes requested):

borrowd_notifications/models.py
```diff
@@ -13,6 +11,9 @@
 )
 from django.urls import reverse
 
+# from .base.models import AbstractNotification
```
vguzman812, line 14: if this is unused, remove it.

borrowd_notifications/signals.py
vguzman812: removing the description field removes the email subject line, so do not remove that.

borrowd_notifications/models.py
```diff
@@ -0,0 +1,401 @@
...
+        "Item requested",
+        "{requester_name} wants to borrow your {item_name}",
+    )
+    ITEM_REQUEST_ACCEPTED = (
+        "Item request accepted",
+        "Your request for {item_name} was accepted by {item_owner_name}",
+    )
+    ITEM_REQUEST_DENIED = (
+        "Item request denied",
+        "Your request for {item_name} was declined",
```
vguzman812: AC 3.2 says "All notification strings must follow the "Actor + Action + Object" format, using the actor's first name".
Please implement this for all notification messages here at the moment.

borrowd_notifications/views.py
Pejjp: added NotificationMetadata and visible_in_app

borrowd_notifications/models.py
Pejjp: Updated all the notifications not following this rule

Pejjp:
Good catches.
I Solved for all raised issues.

borrowd_notifications/migrations/0002_populate.py
vguzman812, line 1: would love to get a second set of eyes on the migrations just to be sure @ajk8 or anyone else.

borrowd_notifications/services.py
```diff
@@ -1,176 +1,297 @@
...
+    @staticmethod
+    def _get_enabled_channels(
+        user: BorrowdUser, notification_type: NotificationType
+    ) -> set[ChannelType]:
+        channels: set[ChannelType] = set()
+
+        if notification_type in NotificationType.mandatory_types():
+            channels.update({ChannelType.APP, ChannelType.EMAIL})
+
+        try:
```
vguzman812: Isn't this try-catch just doing [get_or_create](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#get-or-create) essentially ? 
You could drop this and the except return and just do like:
```python
pref, _ = NotificationPreference.objects.get_or_create(
    user=user,
    notification_type=notification_type.value,
    defaults={
        'in_app_enabled': True, 
        'email_enabled': True, 
        'push_enabled': False
    },
)
```

borrowd_notifications/services.py
```diff
@@ -1,176 +1,297 @@
...
+    @staticmethod
+    def _schedule_summary_digest(recipient: BorrowdUser) -> dict[str, object]:
+        return {
+            "recipient_id": recipient.pk,
+            "scheduled_for": (timezone.now() + _SUMMARY_DIGEST_DELAY).isoformat(),
+            "status": NotificationState.PENDING.value,
+        }
+
+    @classmethod
+    def send_notification(cls, notification: Notification) -> None:
```
vguzman812, line 135: AC 2.4 says notification delivery failure "must never abort the primary user transaction or database action". The current setup here skips any future callbacks if an error arises here. For instance, if a borrow request is rejected, which both sends the user a notification as well as makes the item available again, and an error arises while sending a notification for whatever reason, the second notification to people who want to know if the item is available again is not sent. See https://docs.djangoproject.com/en/5.2/topics/db/transactions/#performing-actions-after-commit. 

A simple fix is just wrapping the whole body in a try-catch here.
```python
 def send_notification(cls, notification: Notification) -> None:
      try:
          if notification.actor == notification.recipient:
              return
      ...
          app_result = payload.data.channels.get(ChannelType.APP)
          NotificationMetadata.objects.update_or_create(
              notification=notification,
              defaults={
                  "visible_in_app": app_result is not None
                  and app_result.status == NotificationState.SUCCESS
              },
          )
       except Exception as exc:
          sentry_sdk.capture_exception(exc)
          logger.exception(
            "Notification dispatch failed (pk=%s)", 
            notification.pk
          )
```

alternatively, for readability, you could extract out your function as such:
```python
  def send_notification(cls, notification: Notification) -> None:
      try:
          cls._dispatch(notification)
      except Exception as exc:
          sentry_sdk.capture_exception(exc)
          logger.exception("Notification dispatch failed (pk=%s)", notification.pk)

  @classmethod
  def _dispatch(cls, notification: Notification) -> None:
      # everything that was in `send_notification` previously eactly as it was.
```

borrowd_notifications/views.py
```diff
@@ -1,3 +1,362 @@
...
+                NotificationType.GROUP_MEMBER_JOINED,
+                "A member joined a group you're part of",
+            ),
+            (NotificationType.GROUP_NEEDS_MODERATOR, "Group needs moderator"),
+            (NotificationType.MEMBERSHIP_PENDING, "New member join request"),
+            (NotificationType.MEMBERSHIP_APPROVED, "Membership approved"),
+        ],
+    },
+    {
+        "name": "Wishlist & Community",
```
vguzman812: this shouldbe reomved until it's actually implemented.

borrowd_notifications/views.py
```diff
@@ -1,3 +1,362 @@
...
+
+    try:
+        ntype = NotificationType(type_value)
+        ChannelType(channel_value)
+    except ValueError:
+        return HttpResponse(status=400)
+
+    if (
+        ntype in NotificationType.mandatory_types()
+        and channel_value != ChannelType.PUSH
```
vguzman812: Any reason we are comparing the string value here rather than parsing the enum like `ChannelType(channel_value) != ChannelType.PUSH` to follow the shape from a few lines above?

templates/notifications/inbox.html
```diff
@@ -0,0 +1,119 @@
...
+                    <button type="submit" class="btn btn-warning btn-sm text-base-content/60">
+                        {% trans "Delete all" %}
+                    </button>
+                </form>
+            {% endif %}
+        </div>
+
+        {% if page_obj.object_list %}
+            <div id="notification-list"
+                 hx-get="{% url 'notification-inbox' %}"
```
vguzman812: If a user is on page 3 and this refresh happens, will notifications from page 1 or page 3 be displayed?

Pejjp: I'll prevent this by triggering the polling only on page one.

borrowd_notifications/services.py
Pejjp: Yes, I guess this is safer. I considered **delivery failures** as the channel failing, so I added try blocks there. but this makes sense.

borrowd_notifications/signals.py
```diff
@@ -3,18 +3,18 @@
...
 call which will create a Notification object for each user in the recipient list. A separate signal handler
-send_notification_email() will catch Notification objects pre-save, send emails based on Notification attributes,
+send_notification() will catch Notification objects post-save, send emails based on Notification attributes,
 and fill in the reserved "emailed" field.
 
 To add a new Notification:
     - add/update a signal handler for the object triggering the notification and call notify.send()
-    - in NotificationService, add a corresponding NotificationType and context in _get_template_context_for
+    - add a corresponding NotificationType and context in NotificationType._get_template_context_for
     - add text and html templates in templates/notifications
```
vguzman812: Missing the need to add the new notification type to `NOTIFICATION_CATEGORIES` and optionally `mandatory_types`

borrowd_notifications/services.py
```diff
@@ -1,176 +1,297 @@
...
+    @classmethod
+    def send_pending_digests(cls) -> int:
+        """Send summary digest emails for all due pending digests. Returns count sent."""
+        now = timezone.now()
+
+        # Notification.data is stored as serialised text (jsonfield library), so
+        # nested ORM key lookups are not supported. Pre-filter by a known substring
+        # then do precise Python-side checks for status and scheduled_for.
+        pending = list(
+            Notification.objects.filter(data__contains="summary_digest").select_related(
```
vguzman812, line 213: This is the same issue as before with checking if the json blob had "APP" to check for in app notifications. If you want to tackle it, feel free, but it's not necessary at this point.

borrowd_notifications/models.py
```diff
@@ -0,0 +1,427 @@
...
+    AvailabilitySubscription,
+    Transaction,
+    TransactionStatus,
+)
+from borrowd_users.models import BorrowdUser
+
+
+class NotificationType(models.TextChoices):
+    # Lending lifecycle
+    ITEM_REQUESTED = (
```
vguzman812: With these values being stored as `notification_email_template_name = ("lookup key", "in-app message template string")`, I worry about someone in the future changing the lookup key slightly because they don't realize it's a db key (e.g., "item notify when available" vs "item notify when available.") This will result in dropped preference rows without a backfill. Also, the second value is a [label](https://docs.djangoproject.com/en/5.2/ref/models/fields/#enumeration-types), and I think it is better to leave this blank and make a separate map for the message templates. 

My recommendation then is to do:
```python
NotificationType(models.Textchoices):
    """ Notifications the app can emit.
    `enum_choice = db_key`
    renaming a db_key requires a migration to backfill the notifictions, otherwise all rows of that type are orphaned.  
    """

    # Lending lifecycle
    ITEM_REQUESTED = "ITEM_REQUESTED"
    ITEM_REQUEST_ACCEPTED = "ITEM_REQUEST_ACCEPTED"
    # everything else
    ...

    def __str__(self) -> str:
        # Drives the email template filename.
        return self.name.lower()

    @property
    def message_template(self) -> str:
        return _MESSAGE_TEMPLATES[self]

    ...
...

# in-app notification message templates
_MESSAGE_TEMPLATES: dict[NotificationType, str] = {
    NotificationType.ITEM_REQUESTED: "{requester_name} wants to borrow your {item_name}",
    NotificationType.ITEM_REQUEST_ACCEPTED: "{item_owner_name} accepted your request for {item_name}",
    ...
    # all the other notification types and what to display in-app
```

borrowd_notifications/signals.py
vguzman812: I would also add in a portion here on where to find all the user-facing copy portions of a notification.
- in-app notification templates
- email subject lines
- email body templates
- preference page labels

vguzman812 (dismissed):
It's generally looking pretty good. Just a few minor aspects to address now versus in a follow-up, with the biggest being the try-catch in services.py and the notification types in models.py

vguzman812 (commented):
I think this is quite good at the moment. The only issue I have is with the summary digest logic/tracking at the moment, but it's not very important at this point. I would feel pretty decent about this being approved.

Thanks for all your hard work @Pejjp

borrowd_notifications/views.py
vguzman812: so you removed this section, but there are still references to community request/wishlist in the migrations and templates and i'm not sure where else. As soon as the community request stuff is removed from this pr, i'm happy to approve this.

Pejjp: We want to keep it for the implementation right? There is no trigger so the only thing to add in the community request or would be the signals. It is also part of the specification.

vguzman812: I see it now. My apologies.

vguzman812 (approved):

ike-sudo404 (approved):

PR #504, opened 2026-06-13
chore: normalize signals.py line endings to LF

vguzman812:
For whatever reason, borrowd_notifications.signals.py escaped the line-ending normalization we did previously. This commit simply changes the line endings from CRLF to LF per .gitattributes. No other changes are included.


leilabl (approved):
No code changes, just whitespace.

PR #505, opened 2026-06-13
Chore/mypy toolchain

vguzman812:
## Summary
Part of a series to fix the various mypy ignore issues we have.

This first PR moves mypy into the project venv so it can see our dependencies, then deletes the pile of suppressions that only existed because the old setup couldn't. Net effect: `# type: ignore` count drops from 126 to 58. Follow-ups will be to enable mypy django stubs and then to deal with third-party typing (django-filter, allauth, guardian, etc).

## Linked Issue(s)

N/A

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [X] Refactor / cleanup
- [ ] Documentation
- [X] Configuration / infrastructure (arguably)

## Changes Made

- `pyproject.toml`, `uv.lock`, `requirements.txt`, `.pre-commit-config.yaml`, `mypy.ini`: 
  - added `mypy` and `django-stubs[compatible-mypy]` to dev dependencies. 
  - Replaced the `mirrors-mypy` pre-commit hook with `uv run mypy .`. 
  - Moved `--strict` and the Python version from the old hook args into `mypy.ini`.
- `mypy.ini`: 
  - removed the `[mypy-guardian.*] ignore_missing_imports` block and sentry-sdk's `import-not-found` since we now read from .venv rather than the dependencies block in the `mirrors-mypy` hook.
- Deleted suppressions that became redundant once mypy could see real dependencies: 
  - many`# type: ignore` comments and `cast()` calls across services/views/signals/models.
- `mypy.ini`, `borrowd/util.py`, `borrowd_groups/views.py`: 
  - enabled `enable_error_code = ignore-without-code`

## How to Test

1. `uv sync`
2. `uvx pre-commit run --all-files`
3. `uv run mypy .`
4. `uv run manage.py test`
5. Add a `# type: ignore` to any line and re-run mypy: it should error with `ignore-without-code`.

## Screenshots (if UI change)

N/A

## Checklist
- [X] I have tested this locally
- [ ] I have checked this works for both moderators and regular members
- [X] No debug logs, or unrelated changes included
- [ ] Migrations are included

## Risk / Notes for Reviewer

N/A


borrowd/util.py
```diff
@@ -148,10 +148,7 @@ def decode(variable: str):  # type: ignore
         JSON decoding error.
     """
     try:
-        if sys.version_info[1] > 5:
```
vguzman812, line 151: This was checking "is the minor version greater than 5," i.e. "is this Python 3.6 or newer". Since Borrow'd uses Python 3.13 (see pyproject.toml), `sys.version_info[1]` is always 13, so the else never runs. 
@ajk8 lmk if there's some other reason this was included, but I see no reason to keep it.

vguzman812: btw the check would also pass Python versions 1.6 + or 2.6 +

requirements.txt
vguzman812, line 1: django stubs and mypy sub-dependencies

bdebut01 (approved):
Image

PR #509, opened 2026-06-16
Feat/push notification channel

### LINKED ISSUE #493: Feature 1: Pluggable Routing Engine, Unified Preferences & Triggers (Backend & Preferences)

Pejjp:
## Summary

Implement push notification channel using `web-push`. 
This PR needs #503 to be merged first. 

## Linked Issue(s)

Closes #493, and is the suite of #503 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [ ] Refactor / cleanup
- [ ] Documentation
- [x] Configuration / infrastructure

## Changes Made

| Area | Changes |
|---|---|
| `GET /sw.js` | serves the service worker from the root origin (required for full-scope push delivery). Source lives at `static/js/sw.js`.
| `POST /push/subscribe/` | saves a new `PushSubscription` for the authenticated 
| `POST /push/unsubscribe/` | removes the subscription by 
| `PUSH Channel implementation` |  Added `PUSH` to `ChannelType`; its label maps to the `push_enabled` field on `NotificationPreference`. - `PUSHNotificationStrategy.send()` sends via `pywebpush.webpush()` (VAPID). Push is never mandatory — all types are opt-in. |
| `PUSH Channel implementation` |  Bulk-toggle for PUSH uses `_all_types_for_scope()` (includes mandatory types) rather than `_optional_types_for_scope()`, since push has no mandatory entries. |

### VAPID configuration
Three new env vars (see `.env.example`):

### Push Compatibility (not all tested)
- Compatible with Chrome, Firefox, Edge (browser and installed PWA).
- Works in Safari 16.4+ for installed PWAs (home screen).

### Add preference-aware notification tests
Cover NotificationService routing logic with 20 new tests across four classes:
| Test Class | Coverage |
|---|---|
| `NotificationPreferenceRoutingTests` | Verifies optional notifications dispatch only to enabled channels and are fully suppressed when no preference row exists; verifies mandatory types bypass preference settings and always reach APP + EMAIL. |
| `NotificationChannelErrorTests` | Verifies per-channel fault isolation: a failing backend records an `ERROR` result without blocking the other channel. |
| `NotificationEmailThrottleTests` | Verifies the hourly email cap drops EMAIL, keeps APP, persists a `summary_digest` entry, and covers email-only throttled paths plus mandatory-type behavior. |
| `NotificationPreferenceToggleViewTests` | Covers `toggle_preference` and `bulk_toggle_preferences` endpoints: row creation, field update, 403 on mandatory types, 400 on invalid inputs, and bulk category disables. |

## How to Test

1. Run uv run manage.py migrate then uv run manage.py runserver
2. Log in as two users (user A and user B) who share a group
3. As user B, request to borrow an item from user A — user A should receive a notification in /notifications/
4. Accept the request as user A — user B receives a notification
5. Walk through collect/return steps and verify both parties receive the appropriate notifications
6. Visit /settings/notifications/ manually - toggle a channel off for a notification type; re-trigger that notification type and confirm it is no longer delivered via that channel
7. Toggle the category-level and master toggles and verify all children switch state instantly without a page reload
8. As a group moderator, approve a pending membership — the new member should receive "Membership approved" and existing members "Change to group membership"
9. Visit /notifications/, dismiss a notification, verify it loses its unread styling; use "Mark all as read" and verify the badge clears

### Stress-testing notifications

50 sends per scenario, 5 push destination(s) per user.

| Scenario | Channels | Mean | Median | P95 | Min | Max |
|---|---:|---:|---:|---:|---:|---:|
| APP only | APP | 4.7ms | 2.5ms | 16.6ms | 2.1ms | 33.2ms |
| EMAIL only | EMAIL | 2.4ms | 2.4ms | 2.9ms | 2.1ms | 2.9ms |
| PUSH only | PUSH | 2.4ms | 2.3ms | 3.0ms | 2.1ms | 3.8ms |
| APP + EMAIL | APP+EMAIL | 2.4ms | 2.3ms | 2.9ms | 2.1ms | 3.3ms |
| APP + PUSH | APP+PUSH | 2.4ms | 2.3ms | 2.7ms | 2.1ms | 2.8ms |
| EMAIL + PUSH | EMAIL+PUSH | 2.4ms | 2.3ms | 2.7ms | 2.1ms | 2.8ms |
| APP + EMAIL + PUSH | APP+EMAIL+PUSH | 2.5ms | 2.5ms | 3.0ms | 2.0ms | 4.1ms |

### Group sends

10 recipients.

| Scenario | Channels | Mean | Median | P95 | Min | Max | Per-recipient mean |
|---|---:|---:|---:|---:|---:|---:|---:|
| Broadcast (no push) | APP+EMAIL | 25.6ms | 24.2ms | 31.7ms | 20.7ms | 67.3ms | 2.6ms |
| Broadcast (all channels) | APP+EMAIL+PUSH | 24.2ms | 23.9ms | 27.7ms | 21.7ms | 31.8ms | 2.4ms |

## Checklist

## Risk / Notes for Reviewer

- Vapid seems to not work on chrome for Mac. Needs tests on other devices and browsers.


borrowd_notifications/push_views.py
```diff
@@ -0,0 +1,59 @@
...
+def vapid_public_key(request: HttpRequest) -> JsonResponse:
+    return JsonResponse({"publicKey": settings.VAPID_PUBLIC_KEY})
+
+
+@login_required
+@require_POST
+def subscribe_push(request: HttpRequest) -> HttpResponse:
+    try:
+        data = json.loads(request.body)
+        endpoint: str = data["endpoint"]
```
vguzman812, line 53: This is a security issue for the app. Since we're storing the endpoint without any validation here, and `borrowd_notificaions/channels.py`'s use of `webpush` in `send` does `requests.post(endpoint, ...)`, any user can make the endpoint something that could access internal port endpoints. For instance, if we were using AWS, a [malicious user could gain instance metadata through this method](https://www.resecurity.com/blog/article/ssrf-to-aws-metadata-exposure-how-attackers-steal-cloud-credentials).

borrowd_notifications/channels.py
```diff
@@ -91,7 +101,61 @@ def send(self, payload: NotificationPayload) -> None:
...
+                    or context.get("group_url")
+                    or f"{base_url}/notifications/"
+                ),
+            }
+        )
+
+        errors: list[str] = []
+        for sub in subscriptions:
+            try:
+                webpush(
```
vguzman812, line 166: We should add an explicit timeout here, especially as the given pushes are not queued.
https://requests.readthedocs.io/en/latest/user/quickstart/#timeouts

vguzman812 (dismissed):
The push notifications (really all notifcations, but push is the important one that is an issue with the current implementation) here are running synchronously on the user's request thread rather than in a bg worker. When a transaction is saved, a Notification post save then occurs for the transaction and `send_notification` then does `transaction.on_commit(...)`. Since Django signals are [always executed in-process](https://stackoverflow.com/a/19731987), and our current path here [uses the synchronous path](https://docs.djangoproject.com/en/6.0/topics/signals/#django.dispatch.Signal.asend), this means the web push holds up the thread until the notification gets sent through. This, plus no timeout on the push, is something that needs to be addressed, or at least talked about. At a minimum, timeout needs to be implemented. This issue also meand the PR violates PRD AC 2.4: "A failure in notification delivery must never abort the primary user transaction".

Push notifications should ideally go into a queue. 
Possible options could be:
- RabbitMQ/redis + celery
- [Django RQ](https://github.com/rq/django-rq) + redis
- django tasks

Arguably these should be set up before this pr even goes through, so I would recommend talking about this and seeing what priorities are, since the PRD does say push notifications are for a v2. If we do switch over to a queue, all of the notifications should get sent through it as well.

Notification process outlined below:


1. A user action hits a view; e.g., accepting a borrow request. That saves a `Transaction`, which fires a `post_save` signal whose receiver calls `notify.send(...)` (`borrowd_notifications/signals.py`, the many `notify.send` calls). Then `notify.send()` creates the `Notification` row.
2. Creating that `Notification` row fires another `post_save`, the `send_notification` receiver (`borrowd_notifications/signals.py:67`):
```python
transaction.on_commit(
    lambda: NotificationService.send_notification(instance), robust=True
)   # signals.py:80-82
```
3. `NotificationService.send_notification` then uses `_dispatch`, and in the channel loop (`services.py:174-177`):
```python
for channel in channels:
    backend = cls._get_backend(channel)
    backend.send(payload)          # blocking call
```
4. For PUSH, `backend.send()` calls `webpush(...)` (`borrowd_notifications/channels.py:142`), which internally does `requests.post(endpoint, timeout=None)`.

borrowd_notifications/push_views.py
Pejjp: The fix I propose is to restrict endpoint to an allowlist of known `push-service hosts: fcm.googleapis.com, updates.push.services.mozilla.com, web.push.apple.com, *.notify.windows.com`, it's the actual SSRF fix, since legitimate browser subscriptions only ever come from that small set of vendor domains and none of our endpoints would contain such patterns.

ajk8 (commented):
Automated review pass (3 independent reviewers, convergent findings). Merge-blockers (broken `.env.example`, `request.user` type-ignore, message_template fallback bug, swallowed PUSH exceptions, account-wide push disable from stale per-browser permission state) already have fixes drafted separately — see follow-up. The comments below cover the remaining should-fix/nitpick findings for your judgment.

borrowd_notifications/channels.py
```diff
@@ -91,7 +101,61 @@ def send(self, payload: NotificationPayload) -> None:
...
+        No native push framework required.
+        """
+        subscriptions = list(
+            PushSubscription.objects.filter(user=payload.notification.recipient)
+        )
+        if not subscriptions:
+            payload.data._error(
+                channel=ChannelType.PUSH, error="No subscription for this device."
+            )
+            return
```
ajk8, line 137: The old stub was intentionally a no-op specifically to avoid a false `SUCCESS` on the audit trail. Now, any user with the PUSH preference enabled but no completed browser subscription (a common transitional state — toggling the preference and completing the browser subscribe flow are independent steps) gets `_error()` on every notification. That's likely to make `ERROR` the *normal* state for a large fraction of users rather than a signal of an actual delivery problem. Worth a product/design call: is this the intended semantics, or should this state be distinguished from a real webpush failure (e.g. a `NOT_SUBSCRIBED` status distinct from `ERROR`)?

borrowd_notifications/channels.py
```diff
@@ -91,7 +101,61 @@ def send(self, payload: NotificationPayload) -> None:
...
+                )
+            except WebPushException as exc:
+                errors.append(str(exc))
+                if exc.response is not None and exc.response.status_code in (404, 410):
+                    sub.delete()
+            except Exception as exc:
+                errors.append(str(exc))
+
+        if errors:
+            payload.data._error(channel=ChannelType.PUSH, error="; ".join(errors))
```
ajk8: If N of a user's devices succeed and only one fails (e.g. one stale/expired subscription), the whole channel is marked `ERROR` even though the user was actually notified on their other devices. Consider only erroring when *all* subscriptions fail, or recording partial success explicitly.

borrowd_notifications/push_views.py
```diff
@@ -0,0 +1,85 @@
...
+        user: BorrowdUser = request.user  # type: ignore[assignment]
+    except (json.JSONDecodeError, KeyError, TypeError):
+        return HttpResponse(status=400)
+
+    if not _is_allowed_push_endpoint(endpoint):
+        return HttpResponse(status=400)
+
+    PushSubscription.objects.update_or_create(
+        endpoint=endpoint,
+        defaults={"user": user, "p256dh": p256dh, "auth": auth},
```
ajk8: `update_or_create(endpoint=endpoint, ...)` has no check that `endpoint` isn't already owned by a different user. Since `endpoint` is globally unique, submitting an endpoint string already registered to someone else silently reassigns that subscription away from its real owner — a quiet loss of push delivery for the original user, with no ownership verification at all. Low likelihood given push endpoint tokens are high-entropy, but worth a defense-in-depth check (e.g. reject/ignore if the existing row belongs to a different user).

borrowd_notifications/push_views.py
```diff
@@ -0,0 +1,85 @@
...
+        auth: str = data["keys"]["auth"]
+        # mypy complains that `request.user` is a AbstractBaseUser or
+        # AnonymousUser, but when I follow the code it looks like it's
+        # AbstractUser or AnonymousUser, which we *would* comply with
+        # here (BorrowdUser subclasses AbstractUser).
+        user: BorrowdUser = request.user  # type: ignore[assignment]
+    except (json.JSONDecodeError, KeyError, TypeError):
+        return HttpResponse(status=400)
+
+    if not _is_allowed_push_endpoint(endpoint):
```
ajk8, line 65: If the client sends a non-string JSON value for `endpoint` (e.g. a number), `data["endpoint"]` succeeds (no `KeyError`/`TypeError`) so it isn't caught by the try/except above, but `_is_allowed_push_endpoint(endpoint)` then calls `urlparse(endpoint)` here, which raises `AttributeError` on a non-str — producing an unhandled 500 instead of a 400. Minor robustness gap, reachable by any authenticated client sending malformed JSON.

static/js/sw.js
```diff
@@ -0,0 +1,16 @@
+self.addEventListener("push", (event) => {
+  const data = event.data ? event.data.json() : {};
+  event.waitUntil(
+    self.registration.showNotification(data.title ?? "Borrow'd", {
+      body: data.body ?? "",
+      icon: data.icon ?? "../icon.svg",
+      badge: '../icon.svg',
```
ajk8: `badge` is unconditionally hardcoded to `'../icon.svg'` (the server never sends a `badge` key to fall back from). This resolves relative to the service worker's own scope (`/sw.js`), not to where the icon actually lives (`/static/icon.svg`), so the notification badge will 404 on every push. Same relative-path issue applies to the `icon` fallback on the line above, though that one is currently unreachable since the server always populates `icon`.

borrowd_notifications/channels.py
```diff
@@ -91,7 +101,61 @@ def send(self, payload: NotificationPayload) -> None:
...
+            body = payload.notification_type.message_template.format(**context)
+        except KeyError:
+            body = payload.notification_type
+
+        base_url = settings.BASE_URL.rstrip("/")
+        push_data = json.dumps(
+            {
+                "title": "Borrow'd",
+                "body": body,
+                "icon": f"{base_url}/static/icon.svg",  # safari doesnt support this.
```
ajk8: Hardcoding `f"{base_url}/static/icon.svg"` duplicates an assumption about `STATIC_URL`/static storage that's otherwise expressed via Django's `{% static %}`/`static()` helpers everywhere else this asset is referenced. Harmless today, but silently breaks with no test signal if static serving config changes (media already moved to S3 in prod).

borrowd_notifications/views.py
```diff
@@ -177,8 +177,11 @@ def _build_preferences_context(user: BorrowdUser) -> dict[str, Any]:
 
 @login_required
 def notification_preferences_view(request: HttpRequest) -> HttpResponse:
+    from django.conf import settings
```
ajk8: `from django.conf import settings` is imported locally inside this function rather than at module top alongside the rest of the file's imports — no apparent circular-import reason, just inconsistent with the file's style.

borrowd_notifications/management/commands/stress_test_notifications.py
```diff
@@ -0,0 +1,310 @@
...
+            user=recipient,
+            notification_type=ntype.value,
+            defaults={
+                "in_app_enabled": app,
+                "email_enabled": email,
+                "push_enabled": push,
+            },
+        )
+
+    def _run_with_patches(self, push: bool, fn: Any) -> list[float]:
```
ajk8: `_run_with_patches` is defined but never called — `_run_scenario`/`_run_group_scenario` each re-implement the same `override_settings`/`patch`/`nullcontext` block inline instead of reusing it. Dead code / missed reuse opportunity.

ajk8:
## Performance follow-up: push notification fan-out

Flagging a performance/reliability risk in this PR's synchronous PUSH dispatch, plus what's been done about it and what's still open.

**Landed in `11dad04`:** `webpush()` was being called with no timeout — `pywebpush`'s default is `timeout=None`, which passes straight through to `requests.post(timeout=None)`, i.e. **no timeout at all**. On this app's 2-worker synchronous gunicorn deployment (`gunicorn -w 2`, no `--timeout` override → 30s default), an unresponsive push vendor could tie up half of total request capacity indefinitely. Now bounded to 5s per device.

**Why this matters beyond the single-call case:** notification fan-out for a group broadcast (e.g. `GROUP_MEMBER_JOINED`) creates one `Notification` row per active member, and each row's `post_save` → `transaction.on_commit` runs `NotificationService._dispatch()` synchronously, in order, all within the same HTTP request — before the response returns. For each recipient with PUSH enabled, that's one blocking `webpush()` call per device. For a group of size N with D devices/member, that's `N × D` sequential blocking calls in one request. Most groups are small, but occasional groups here are expected to exceed 50 members — at even 100-300ms/call (realistic push-relay latency; not something this app has measured in production yet), that's already 5-30+ seconds sequentially, i.e. at or past gunicorn's 30s worker-kill threshold under **normal, healthy** conditions, before factoring in the now-bounded-but-still-real per-call timeout.

Given the "avoid new infrastructure for now" constraint (no Redis/Celery), two fast-follow mitigations, in order of leverage:

1. **Parallelize per-device/per-recipient `webpush()` calls with a bounded `concurrent.futures.ThreadPoolExecutor`** (stdlib only). Collapses N sequential blocking calls into a handful of concurrent round trips — cuts large-group worst-case wall-clock by ~5-10x. Still occupies the gunicorn worker for the (much shorter) duration, so this helps request latency but not the "worker unavailable to serve other requests" risk during that window.

2. **Move push sends off the request path using the outbox/cron pattern already established in this codebase.** The EMAIL channel already defers via a `summary_digest` entry drained by the `send-summary-digests` cron (`.platform.app.yaml`, every 15 min). Extending the same idea to PUSH — write pending sends to a table instead of sending inline, drain it via a new, more frequent cron entry — fully decouples push latency from user-facing requests using only what's already provisioned (Postgres + platform.sh cron, no new paid services). This is the structural fix; #1 is a stopgap that reduces the blast radius in the meantime.

Not asking to block this PR further on these — landing them as separate fast-follows once #1 (timeout) is in seems reasonable given current traffic is very low, but worth prioritizing before this scales to "tens of concurrent requests."

borrowd_notifications/push_views.py
Pejjp: added UniqueConstraint(fields=["user", "endpoint"], name="unique_push_subscription_per_user") in Meta.

borrowd_notifications/push_views.py
Pejjp: Added a type check on subscription

borrowd_notifications/channels.py
Pejjp: Added the `NOT_SUBSCRIBED` status, and a cta appears on the profile page for the user to enable browser permissions

borrowd_notifications/channels.py
Pejjp: added partial success recording

static/js/sw.js
Pejjp: resolved

borrowd_notifications/channels.py
Pejjp: added

ajk8:
## Review

Solid, security-conscious implementation of the PUSH channel. The SSRF allow-list on `subscribe_push` (only known vendor push hosts, not a block-list) is well designed and well tested, per-device push failures are correctly isolated from the whole channel, and the bounded `webpush()` timeout avoids tying up a sync gunicorn worker. Test coverage is strong across the allow-list edge cases, partial/total failure handling, and the `NOT_SUBSCRIBED` vs `ERROR` distinction.

### Pushed a follow-up commit (`a6dd91a`) fixing two merge-blockers:

1. **Missing VAPID config guard** — `NotificationPreference`'s own "add a new channel" checklist says *"For channels that require external credentials (like PUSH), guard the `send()` call with a settings check"*, but `PUSHNotificationStrategy.send()` didn't do this. Without it, an unconfigured deployment (or a misconfigured one where the private key gets dropped after subscriptions already exist) would hit `pywebpush` on every send and report a generic `ERROR` + a Sentry event per notification instead of failing quietly and audibly once. Added the guard plus a test (`test_missing_vapid_keys_skips_delivery_without_calling_webpush`).
2. **Undocumented env vars** — `VAPID_PRIVATE_KEY`/`VAPID_PUBLIC_KEY`/`VAPID_ADMIN_EMAIL` weren't listed in `docs/CONTRIBUTING.md`, which AGENTS.md points to as the source of truth for env vars. Added entries alongside the others, including the keypair-generation command.

Full test suite (362 tests), `ruff check`/`format`, and `mypy --strict` all pass after these changes.

### Non-blocking follow-ups (not fixed, for future cleanup)
- `PushSubscription`'s explicit field type annotations (`user: ForeignKey[BorrowdUser] = ForeignKey(...)`, etc.) go against the "don't annotate model fields" convention — though this mirrors an existing pattern already in `NotificationPreference` in the same file, so it's pre-existing, not new.
- `PushSubscription` intentionally skips the audit/soft-delete fields (`created_by`, `deleted_at`, etc.) and `unsubscribe_push` does a hard `.delete()` — confirmed with the team this is the right call for a device-token table, not an oversight.
- `_urlBase64ToUint8Array` and the subscribe/permission-check flow are duplicated between `profile.html` and `preferences.html` — worth extracting to a shared static JS module if push-related UI grows.
- `service_worker()` reads `sw.js` from disk on every request with no cache headers — low traffic, but an easy win later.

### Deployment note
Please confirm the three VAPID env vars are set in the prod/cert platform.sh environments before merge — otherwise push silently no-ops there (which is now the correct/guarded behavior, but still worth confirming isn't accidental).

Pejjp (commented):
LGTM !

ike-sudo404 (approved):

PR #510, opened 2026-06-21
added upload loading state and tests

### LINKED ISSUE #475: [Bug]: When photo is "pending upload status"  and user clicks save changes -  they are redirected to the page with error message "Oh no! Something went wrong"
## Summary
When the user tries to save changes having a photo in "pending upload" status, changes are not saved, user sees "Oh no! Something went wrong" page

## Severity
Medium 

## Expected Behavior
1. We can disable the button till the upload process is over
2. We can display a notification of the kind "we could not upload [number] of images. Please try again."
Up to the product team to decide

 ## Actual Behavior
On the edit step, the file is displayed as a broken image and shows "pending upload."
after clicking save changes, user sees "Oh no! Something went wrong" page

## Who Is Affected?
all users

## Steps to Reproduce
1.  Log in to the app
2. Navigate to the Inventory page
3. Navigate to item details page
4. Click edit
5. Choose file from the library
6. Make sure that this file is in pending upload state 
7. Click save changes

Image
Image

## Environment
I faced this problem only using my android phone, more details can be found here https://borrowd.sentry.io/issues/7487340588/?environment=staging&project=4510660596137984&query=&referrer=issue-stream

Pejjp:
## Summary
Fixes item editing with pending photo uploads by validating images before processing and adds an upward shimmer effect while pending images are uploaded.

### small improvement proposal
When the 5 images threshold is reached, the user have to delete manually and upload new images: 
- We could simply add a replace buttons on the image and handle the deleted implicitly.

## Linked Issue(s)
Closes #475 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation

## Changes Made
- Updated `ItemUpdateView` to validate uploaded photos through `ItemPhotoForm`, preventing invalid images from causing a 500 error.
- Added a per-image shimmer animation shown while pending photos are submitted.
- Added regression coverage for corrupt pending image uploads.

## How to Test
1. Edit an item and select one or more valid photos.
2. Save the item and confirm the shimmer appears while submitting.
3. Confirm the item changes and photos are saved.
4. Submit an invalid or corrupt image and confirm the item changes save without displaying the 500 page.

## Screenshots (if UI change)
Add a screenshot or recording of the pending-image shimmer during save.

https://github.com/user-attachments/assets/f6273a61-fbf4-4b0c-9d72-cd4945cbed1e

## Checklist

## Risk / Notes for Reviewer
No model changes or migrations are required. Review the image validation behavior and pending-thumbnail shimmer timing.

### DIFF
--- a/borrowd_items/tests/test_photo_size_validation.py
+++ b/borrowd_items/tests/test_photo_size_validation.py
@@ -12,7 +12,7 @@ from io import BytesIO
 from typing import Any, cast
 
 from django.core.files.uploadedfile import SimpleUploadedFile, UploadedFile
-from django.test import TestCase
+from django.test import RequestFactory, TestCase
 from django.utils.datastructures import MultiValueDict
 from PIL import Image
 
@@ -25,6 +25,7 @@ from borrowd_items.forms import (
     validate_image_size,
 )
 from borrowd_items.models import Item, ItemCategory
+from borrowd_items.views import ItemUpdateView
 from borrowd_users.models import BorrowdUser
 
 
@@ -66,6 +67,40 @@ def make_files(
 def make_empty_files() -> MultiValueDict[str, UploadedFile]:
     """Create an empty files dict for form testing."""
     return cast(MultiValueDict[str, UploadedFile], {})
+
+
+class ItemUpdatePhotoProcessingTests(TestCase):
+    def test_corrupt_pending_photo_is_skipped_without_raising(self) -> None:
+        owner = BorrowdUser.objects.create_user(
+            username="owner",
+            email="owner@example.com",
+            password="password",
+        )
+        item = Item.objects.create(
+            name="Test item",
+            description="Test description",
+            owner=owner,
+            created_by=owner,
+            updated_by=owner,
+            trust_level_required=TrustLevel.STANDARD,
+        )
+        corrupt_photo = SimpleUploadedFile(
+            name="corrupt.jpg",
+            content=b"not an image",
+            content_type="image/jpeg",
+        )
+        request = RequestFactory().post(
+            "/items/1/edit/",
+            data={"new_photos": corrupt_photo},
+        )
+        request.user = owner
+        view = ItemUpdateView()
+        view.request = request
+        view.object = item
+
+        view._process_uploaded_photos()
+
+        self.assertEqual(item.photos.count(), 0)
 
 
 class ValidateImageSizeFunctionTests(TestCase):
--- a/borrowd_items/views.py
+++ b/borrowd_items/views.py
@@ -3,11 +3,12 @@ from urllib.parse import urlencode
 
 from django.contrib import messages
 from django.contrib.messages.api import MessageFailure
-from django.core.validators import FileExtensionValidator
+from django.core.files.uploadedfile import UploadedFile
 from django.forms import ModelForm
 from django.http import HttpRequest, HttpResponse
 from django.shortcuts import get_object_or_404, redirect
 from django.urls import reverse, reverse_lazy
+from django.utils.datastructures import MultiValueDict
 from django.views.decorators.http import require_POST
 from django.views.generic import CreateView, DeleteView, DetailView, UpdateView
 from django_filters.views import FilterView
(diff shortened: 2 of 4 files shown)


borrowd_items/views.py
```diff
@@ -370,24 +369,19 @@ def _process_uploaded_photos(self) -> None:
...
-            allowed_extensions=ALLOWED_IMAGE_EXTENSIONS
-        )
         skipped = 0
 
         for upload in uploaded_files[:remaining_slots]:
-            try:
-                ext_validator(upload)
-                validate_image_size(upload)
-            except Exception:
+            photo_files = MultiValueDict[str, UploadedFile]({"image": [upload]})
```
ajk8: **Confusing constructor syntax** — `MultiValueDict[str, UploadedFile](...)` calls the parameterized generic alias as a constructor. It works in Python 3.9+ (the alias delegates to `MultiValueDict.__init__`), but it reads like a cast/annotation being called and will surprise most readers.

Conventional form:
```python
photo_files: MultiValueDict[str, UploadedFile] = MultiValueDict({"image": [upload]})
```

borrowd_items/tests/test_photo_size_validation.py
```diff
@@ -68,6 +69,40 @@ def make_empty_files() -> MultiValueDict[str, UploadedFile]:
...
+        request = RequestFactory().post(
+            "/items/1/edit/",
+            data={"new_photos": corrupt_photo},
+        )
+        request.user = owner
+        view = ItemUpdateView()
+        view.request = request
+        view.object = item
+
+        view._process_uploaded_photos()
```
ajk8, line 101: **Testing a private method directly** — this test calls `_process_uploaded_photos()` after manually wiring up `view.request` and `view.object`, bypassing Django's view dispatch and any setup that `dispatch()` / `form_valid()` would normally do. It couples the test to the private method name and its exact preconditions.

A more robust approach is to POST through the full Django test client:
```python
self.client.force_login(owner)
response = self.client.post(
    reverse("item-edit", kwargs={"pk": item.pk}),
    data={"new_photos": corrupt_photo, "name": item.name, ...},
)
self.assertEqual(item.photos.count(), 0)
```
This survives refactors of the private method and catches regressions from view-layer concerns (permissions, form handling, etc.).

templates/items/item_form.html
```diff
@@ -292,4 +297,27 @@
 
     {% include "components/modal.html" with modal_id="trust-level-info-modal" modal_title="Trust levels" modal_size="full" modal_content_template="items/_trust_level_info_content.html" cancel_button_text="Got it" %}
 
+    <style>
```
ajk8: **Inline `<style>` block bypasses the Vite/Tailwind pipeline** — this `<style>` tag renders mid-`<body>` inside a Django template block, outside the build pipeline. The project's other styles flow through Vite (purging, versioning, cache-busting). Placing the keyframe animation here means it can't be tree-shaken and sits apart from all other visual definitions.

Preferred: add a custom keyframe and utility to the project's main CSS file (wherever other custom Tailwind utilities live), then reference it as a class in the template. That keeps all visual logic in one place and in the Vite build.

ajk8 (commented):
Three findings below — one style nit, one testing-altitude issue, and one CSS-pipeline inconsistency. The core validation refactor (switching from manual validators to `ItemPhotoForm`) is correct: extensions, size, and now Pillow content-validity are all enforced equivalently or more strictly than before. No functional regressions found.

borrowd_items/tests/test_photo_size_validation.py
ajk8: This is not a required change.

borrowd_items/views.py
ajk8: I agree this is confusing, but no need to block the PR on it.

templates/items/item_form.html
ajk8: This should probably change.

Pejjp: Moved the style to main.css.

leilabl (approved):

borrowd_items/tests/test_photo_size_validation.py
Pejjp: This test is just focused on testing the processing of a corrupted image, so I didn't se the need to use the Django's view.

PR #513, opened 2026-06-23
Refactor/mypy django plugin

vguzman812:
## Summary
Part of a series to fix the various mypy ignore issues we have.
This second PR enables the django-stubs mypy plugin, which fixes things like FK reads, reverse accessors, `<fk>_id` attrs, etc. 
`# type: ignore` count drops from 58 to 42. 
The follow-up will be third-party typing (django-filter, allauth, guardian, etc).

## Linked Issue(s)

N/A

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [X] Refactor / cleanup
- [ ] Documentation
- [X] Configuration / infrastructure (arguably)

## Changes Made

- `mypy.ini`, `.github/workflows/ci.yml`:
  - enabled `plugins = mypy_django_plugin.main`
  - CI pre-commit job now exports `DJANGO_SECRET_KEY` because the plugin imports the settings module, and that requires the env var to be present.
- `pyproject.toml`, `uv.lock`, `requirements.txt`:
  - added `django-stubs-ext` as a runtime dependency.
- `borrowd/config/base.py`:
  - replaced the hand-rolled `__class_getitem__` monkeypatch loop with `django_stubs_ext.monkeypatch()`.
- various files:
  - stripped the explicit field annotations and `#type: ignore[attr-defined]`. With the plugin, a single-arg `ForeignKey[X]` annotation overrides inference and leaves the read type unresolved, which is what forced the old `attr-defined` ignores in the first place.
- `borrowd_groups/models.py`:
  - renamed `BorrowdGroupManager.create` to `create_group` (the plugin hard-checks any manager method named `create` against model fields, and `trust_level` is deliberately not a field). Call sites updated (all in tests).
- `borrowd_groups/mixins.py`:
  - deleted the whole file as it only contains `BorrowdGroupPermissionMixin`, which was noted as "pre-release functionality"; 
    -  it was used in borrowd_users/models.py: `class BorrowdUser(AbstractUser, BorrowdGroupPermissionMixin, GuardianUserMixin)`.
      - `AbstractUser` already has a `groups` field, so the permission mixin is never used.
- `borrowd_beta/management/commands/generate_code.py`:
  - The plugin flagged a `BetaCode.objects.create(...)` call passing a nonexistent `email` field and omitting the required audit FKs; fixed to use `get_system_user()`.
- `borrowd_users/request.py` (new):
  - The plugin flagged 8 sites for having incompatible types for users when we had `request.user` as `BorrowdUser | AnonymousUser` since `AnonymousUser` isn't a valid value for a FK to `BorrowdUser`. Therefore, I added `get_authenticated_user()` to fix this. I only did these 8 sites, but there are quite a few others, and PR 3 gets the rest.
  - affected files: `borrowd_beta/admin.py`, 
- `AGENTS.md`: 
  - Updated the typing conventions to match (no more handwritten field annotations).
- `borrowd_groups/signals.py`:
  - `perms_group` can be `None`, but two signal handlers read `.name`/`.pk` off it without checking (a latent `AttributeError`). The plugin caught it; added the missing `None` guards.

## How to Test

1. `uv sync`
2. `uvx pre-commit run --all-files`
3. `uv run mypy .`
4. `uv run manage.py test`
5. `uv run manage.py makemigrations --check --dry-run` should report no changes (for the mixin drop)

## Screenshots (if UI change)

N/A

## Checklist
- [X] I have tested this locally
- [X] I have added/updated relevant tests
- [ ] I have checked this works for both moderators and regular members
- [X] No debug logs, or unrelated changes included
- [ ] Migrations are included

## Risk / Notes for Reviewer

N/A


bdebut01 (changes requested):
Merge https://github.com/borrowd/borrowd/pull/514 (now approved) → merge main to this guy to fix the e2e fail on here and i'm all good to approve this one!

vguzman812:
> Merge #514 (now approved) → merge main to this guy to fix the e2e fail on here and i'm all good to approve this one!

merged #514 to this branch.

bdebut01 (approved):

vguzman812:
needs an extra approval since @bdebut01 's review was from before I merged the e2e here.

ajk8 (approved):
I love this!

PR #522, opened 2026-07-16
Fix/e2e tests

vguzman812:
## Linked Issue(s)

Closes #

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation

## Changes Made

-
-

## How to Test

1.
2.

## Checklist

### DIFF
--- a/e2e_tests/pages/profile_page.py
+++ b/e2e_tests/pages/profile_page.py
@@ -7,6 +7,9 @@ class ProfilePage:
     def __init__(self, page: Page):
         self.page = page
         self.heading = page.get_by_role("heading", name="Profile")
+        self.personal_info_toggle = page.locator(
+            "summary", has_text="Personal information"
+        )
         self.first_name_input = page.locator("#id_first_name")
         self.last_name_input = page.locator("#id_last_name")
         self.email_input = page.locator("#id_email")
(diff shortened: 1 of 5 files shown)


e2e_tests/pages/profile_page.py
```diff
@@ -16,6 +19,20 @@ def __init__(self, page: Page):
...
+
+    def open_personal_info(self):
+        """The personal info fields live in a drawer that starts collapsed.
+
+        The drawer is Alpine-driven (CDN script, deferred). A click before
+        Alpine initializes toggles the native <details> without setting
+        openDrawer, leaving the two permanently out of sync until the next
+        reload. Wait for Alpine to claim the drawer before clicking.
+        """
+        self.page.wait_for_function(
```
Pejjp, line 32: Not blocking, it passes today but worth hardening before this wait pattern gets copied to other drawers/modals in future tests.

`document.querySelector('[x-data]')` returns the first match in document order, which is `<body x-data> `in `templates/base.html:26`, not this page's own `<div x-data="{ openDrawer: null }">` that actually drives the drawer `(templates/users/profile.html:11)`. So this checks whether Alpine has claimed `<body>`, not whether the drawer's own component state is ready. It happens to work today because Alpine initializes the whole tree synchronously in one pass, but that's incidental to Alpine's current implementation, not something this check actually guarantee, and `_x_dataStack` is a private, undocumented internal, not part of Alpine's public API.

Suggest waiting on Alpine's own alpine:initialized event instead (dispatched on document once the whole tree is processed, public API, no version-fragility):

```
self.page.wait_for_function("() => window.__alpineInitialized === true")
<!-- once, e.g. in base.html -->
<script>document.addEventListener('alpine:initialized', () => window.__alpineInitialized = true)</script>
```

Pejjp (approved):
LGTM !

e2e_tests/pages/profile_page.py
bdebut01: This strikes me as a good change, but to keep things rolling Jeff & I agreed we'd merge this PR and toss this change into a new one

bdebut01 (approved):

PR #523, opened 2026-07-17
Feat/507/per item group sharing

### LINKED ISSUE #507: [Feature]: Remove Trust Levels & Introduce Per-Group Item Sharing
PRD Link: [Feature #507 - Remove Trust Levels & Introduce Per-Group Item Sharing](https://docs.google.com/document/d/11ftI9o9sIgvcdgMj45dpKChGkupWEhBoKqJmUFATBBA/edit?usp=sharing)
Design Reference:

---

## Concept Note

Borrow'd currently relies on a "Trust Level" system (Standard vs. High trust) to gate item visibility between group members. Feedback from our community pilots indicates this concept is a major source of user confusion and friction. Users don't think in terms of abstract, tiered trust levels; they think in terms of **contextual sharing** (e.g., *"I want to share my sewing machine with my sewing group, not my biking group"*).

This feature removes the abstract Trust Level system entirely across groups, items, and onboarding. In its place, we will introduce **per-group item sharing**. Users will directly choose which of their groups can see each item via a multi-select selector (defaulting to "All Groups", which automatically propagates to future groups they join).

---

## Problem Statement

The current trust level system introduces three major issues:

1. **High Cognitive Load & Friction:** Gating sharing via an abstract hierarchy (Standard vs. High) forces users to configure a trust tier per group membership and match it against a trust tier per item. This threshold-based model (`item.trust_level_required <= membership.trust_level`) is highly unintuitive.
2. **Onboarding Abandonment:** Because trust levels are complex, onboarding requires a dedicated Step 3 explanation screen, increasing the path length and driving drop-offs before users see the core product value.
3. **Imprecise Control:** A user with a sewing group and a biking group cannot share a sewing machine with *only* the sewing group. The sharing is tier-based, not group-based — the wrong abstraction for the problem.
4. **Ongoing Confusion:** Trust level editors appear on group detail pages, trust level badges on items, and trust level selectors on group create/join flows — each adding cognitive load without proportional value.

---

### Story 1: Choosing Where to Share (Owner — Item Form)

*As an item owner, I want to explicitly select which groups can see my item during creation and editing, so that I maintain precise control over my inventory visibility.*

- **AC1 (Default State):** By default, all newly created items must have the "All Groups" option enabled.
- **AC2 (Implicit Expansion):** If "All Groups" is selected, the item is marked as "shared with all current and future groups." When the owner joins a new group in the future, the item must automatically become visible to members of that new group.
- **AC3 (Mirroring Categories UI):** The selector must use a multi-select modal mirroring the existing Categories filter modal. Selecting "All Groups" disables/clears individual group selections; selecting any individual group deselects "All Groups."
- **AC4 (Visibility Constraint):** The list of groups in the modal must only display active groups where the user is currently an approved member.
- **AC5 (Zero Groups Guard):** If the user has no active group memberships, the modal shows "All Groups" as the only option (greyed-out chips area), and the item is saved with `share_with_all_groups = True` by default.
- **AC6 (Responsive Design):** The modal must render cleanly and be fully interactable on both mobile screens and desktop.

### Story 2: Viewing Sharing Status (Owner — Item Detail)

*As an item owner, I want to see which groups my item is shared with at a glance, without navigating into edit mode.*

- **AC1 (Owner View):** The item detail page shows a "Shared with" section displaying either "All Groups" or the specific group names.
- **AC2 (Non-Owner Exclusion):** Non-owners do not see the "Shared with" section.
- **AC3 (Responsive):** Badges wrap cleanly on mobile.

### Story 3: Group Join Propagation

*As a user joining a new group, I want my "All Groups" items to be instantly visible to the new group, without updating each item manually.*

- **AC1 (Instant Visibility):** When a membership status changes to `ACTIVE` (immediately on public group join, or after moderator approval), all items with `share_with_all_groups = True` must automatically grant VIEW permissions to the new group.
- **AC2 (Scoped Isolation):** Items with `share_with_all_groups = False` must **not** be visible to the new group unless that group is explicitly in `shared_with_groups`.

### Story 4: Simplified Group Flows (Group Create, Join, Detail, List)

*As a user, I want group pages free from confusing trust level controls.*

- **AC1 (Create Form):** The trust level selection field and info modal are removed from the Group Create form.
- **AC2 (Join Page):** The trust level selection dropdown and info modal are removed from the Group Join page.
- **AC3 (Detail Page):** The trust level editor component (`<c-groups.trust-level-editor>`) and info modal are removed from the Group Detail page. The page layout adjusts cleanly.
- **AC4 (List / Cards):** The "Group & trust level" header is updated to "Group". The trust level badge is removed from group list cards.
- **AC5 (All Surfaces):** All changes work on both mobile and desktop.

### Story 5: Streamlined Onboarding

*As a new user, I want a short and clear onboarding experience so I can start using the app immediately without wading through permission rules.*

- **AC1 (Step 3 Removal):** Onboarding Step 3 (Trust Levels) is completely removed. Navigating to `/onboarding/3/` returns a 404 or redirects.
- **AC2 (Step 2 Updated):** Step 2 (Items) progress bar is updated to reflect it is now the final step. A brief explainer is added: *"Shared with: You choose which groups see each individual item, so you're always in control."*
- **AC3 (Step 1 Updated):** Step 1 (Groups) progress bar is updated for the 2-step flow.
- **AC4 (Flow Completion):** Clicking "Next" / "Get Started" on Step 2 redirects directly to `onboarding_complete`.
- **AC5 (Responsive):** Updated steps render cleanly on both mobile and desktop.

---

## Flowchart of Journey

```mermaid
flowchart TD
    A["User creates or edits Item"] --> B["Open Group Selector Modal"]
    B --> C{"Select 'All Groups'?"}

    C -- Yes --> D["Disable individual group chips"]
    D --> E["Set share_with_all_groups = True"]
    E --> F["Item visible to all current + future groups"]

    C -- No --> G["Enable individual group chips"]
    G --> H["Select specific subset of groups"]
    H --> I["Set share_with_all_groups = False"]
    I --> J["Populate shared_with_groups M2M"]
    J --> K["Item visible ONLY to selected groups"]

    L["User joins a new Group"] --> M{"Membership ACTIVE?"}
    M -- Yes --> N["Fetch all owner items where share_with_all_groups == True"]
    N --> O["Grant VIEW permissions to new Group"]
    M -- No / Pending --> P["Wait for Moderator approval — no permissions granted yet"]
```

---

### What We're Building

Replace the trust level system with **explicit per-group item sharing**:

1. Each item gets a "Shared with" control — a multi-select group picker mirroring the existing Categories modal pattern
2. "All Groups" is selected by default (applies to future groups the user joins)
3. Users can narrow sharing to specific groups at any time
4. Trust levels are removed from everywhere — groups, items, onboarding, and all related UI

### Why This Solution

- **Mirrors the user mental model** — "Share with Group X" is instantly understandable
- **Reuses a proven UI pattern** — The Categories modal (multi-select with "All" option) already exists and is understood
- **Simplifies onboarding** — Step 3 is removed entirely; Step 2 gets a brief "Shared with" mention
- **Cleaner data model** — Explicit M2M relationship vs. implicit threshold math

---

## Solution Breakdown

The work is broken into 6 components:

1. **Database & Model Changes** — New fields, remove old fields, data migration
2. **Permission Logic Rewrite** — Signals that control item visibility
3. **Item UI** — Form and detail page (group selector replaces trust level)
4. **Group UI** — Remove trust level from create, join, detail, and list pages
5. **Onboarding** — Remove Step 3, update Step 2 and Step 1
6. **Cleanup** — Remove dead components, templates, views, URLs, forms, and tests

---

### 1. Database Schema Updates

**New fields on `Item`:**

| Field | Type | Default | Description |
| --- | --- | --- | --- |
| `share_with_all_groups` | `BooleanField` | `True` | When True, item is visible to all current AND future groups |
| `shared_with_groups` | `ManyToManyField → BorrowdGroup` | empty | Specific groups this item is shared with (used when `share_with_all_groups` is False) |

**Fields to remove:**

| Model | Field | Current Purpose |
| --- | --- | --- |
| `Item` | `trust_level_required` | Minimum trust threshold for visibility |
| `Membership` | `trust_level` | User's trust level in a specific group |

**Data Migration Plan:**

- All existing items with `trust_level_required = STANDARD (1) or HIGH (2)` → set `share_with_all_groups = True`
- Populate `shared_with_groups` with the owner's groups

### Item Form (Create & Edit)

**Remove:** `trust_level_required` dropdown and its info modal

**Add:** "Shared with" group selector — multi-select modal mirroring the Categories modal:

- Button label shows current selection: "All Groups" / "2 Groups" / specific group names
- Modal opens with toggleable group chips
- "All Groups" chip at the top, selected by default
- Mutually exclusive with individual group selection

### Item Detail Page (Inventory View)

**Remove:** High Trust / Standard Trust badge and trust level info modal

**Add:** "Shared with" display — visible to owner only:

- `share_with_all_groups = True` → "All Groups" badge
- `share_with_all_groups = False` → list of specific group name badges

### 3. Group UI Changes

**Group Create Form:** Remove trust level field + info modal
**Group Join Page:** Remove trust level dropdown + info modal
**Group Detail Page:** Remove `<c-groups.trust-level-editor>` component + info modal
**Group List / Cards:** Remove trust level badge; update "Group & trust level" header to "Group"

### 4. Onboarding Changes

**Remove entirely:**

- `onboarding/step3.html` template
- `onboarding_step3` view (`borrowd_web/views.py`)
- URL pattern (`borrowd_web/urls.py`)
- Static asset: `Onboarding Step 3 - Trust Levels.PNG`

**Update Step 2 (Items) → now the final step:**

- Progress bar updated for 2-step flow
- Add copy: *"Shared with: You choose which groups see each individual item, so you're always in control."*
- "Next" / "Get Started" button links directly to `onboarding_complete`

**Update Step 1 (Groups):**

- Progress bar updated for 2-step flow (no content changes)

### Templates to Remove

| Template | Purpose |
| --- | --- |
| `components/trust_level_badge.html` | "High Trust" / "Standard Trust" badge |
| `components/trust_pill.html` | Trust level pill display |
| `components/groups/trust_level_editor.html` | Alpine.js trust level dropdown editor |
| `groups/_trust_level_info_modal_content.html` | Trust level explanation modal (groups) |
| `items/_trust_level_info_content.html` | Trust level explanation modal (items) |
| `onboarding/step3.html` | Trust levels onboarding step |

### Views / URLs to Remove

| Item | Location |
| --- | --- |
| `UpdateTrustLevelView` | `borrowd_groups/views.py` |
| `update-trust-level` URL | `borrowd_groups/urls.py` |
| `onboarding_step3` view | `borrowd_web/views.py` |
| `onboarding_step3` URL | `borrowd_web/urls.py` |

### Forms to Update

| Form | Change |
| --- | --- |
| `GroupCreateForm` | Remove `trust_level` field |
| `GroupJoinForm` | Remove `trust_level` field |
| `UpdateTrustLevelForm` | Delete entirely |
| `ItemForm` | Remove `trust_level_required`, add `share_with_all_groups` + `shared_with_groups` |

### Notification Templates to Update

| Template | Change |
| --- | --- |
| `notifications/group_member_joined.html` | Remove "update your trust level for this group" copy |
| `notifications/group_member_joined.txt` | Same — remove trust level reference |

---

## Areas of Impact

| Area | Impact | Handling |
| --- | --- | --- |
| Item visibility / permissions | Core change — threshold logic → explicit group selection | Rewrite signals, full test coverage |
| Group creation flow | Simplified — no trust level selection | Remove field from form and template |
| Group join flow | Simplified — no trust level selection | Remove field; auto-share on join |
| Group detail page | Simplified — no trust level editor | Remove component |
| Group list / cards | Minor — badge and column header removed | Update template |
| Onboarding | Shortened from 3 steps to 2 | Remove Step 3; update Step 2 copy + progress |
| Item form | Modified — trust level replaced with group selector | New modal component |
| Item detail | Modified — trust badge replaced with "Shared with" display | New component |
| Notifications | Minor copy change | Remove trust level reference |
| Tests | ~15+ test files reference trust levels | Rewrite to use new sharing model |
| Data migration | All existing items need migration | Default to "All Groups", preserve High Trust intent |

---

## Summary

This PRD removes the Trust Level system (Standard / High) and replaces it with explicit per-group item sharing. Users select which groups see each item via a multi-select modal (defaulting to "All Groups"), mirroring the existing Categories modal pattern. Joining a new group automatically propagates visibility for "All Groups" items.

The change simplifies the mental model, removes an entire onboarding step, and eliminates a pervasive source of user confusion — while preserving full, granular control over item visibility.

**Scope:** ~30+ files affected across models, views, forms, signals, templates, and tests. One data migration required (with a safe rollback path). No external API or third-party dependencies.

leilabl:
## Summary
Removes the concept of Trust Level throughout the app. Add item per group sharing instead. When adding or editing an item, the user has the opportunity to share that item with all groups or with specific groups they are a member of.

## Linked Issue(s)
Closes #507 

## Type of Change
- [ ] Bug fix
- [x] New feature
- [x] Refactor / cleanup
- [x] Documentation

## Changes Made

- Removed the TrustLevel model and create migrations to alter item and membership tables
- Removed all the references to trust level in the group files and templates
- Added modal for picking which groups to share item with in the item components
- Update all tests

## How to Test

1. Create a group and verify there is no trust level picker
2. In the groups list there is no trust level badge
3. In group detail page, when you're a moderator, there is no trust level edit picker
4. Create an item and verify there is no trust level picker
5. Create an item and verify there is a group sharing picker listing all the groups you are member of and the option to share with all groups
6. Verify with a secondary user that the visibility of items matches the group membership/item sharing settings
7. Edit an item and verify there is no trust level picker
8. Onboard a new user and verify that there are no references to trust level. 
9. Onboarding has only 2 steps instead of 3.

## Checklist


borrowd.dbml
```diff
@@ -22,7 +16,7 @@ Table user_preferences {
 }
 
 // OPEN would mean that all of your items are shared according to each item's
-// trust_level; CLOSED would mean that you need to individually assign items
```
leilabl, line 25: I'm not sure what CLOSED stands for here. Just replaced trust_level with group_sharing, but I don't think this makes sense.

Interested to know what is the role of this file @team

borrowd_groups/signals.py
```diff
@@ -189,20 +181,8 @@ def refresh_permissions_on_membership_update(
...
-        )
-
-        for item_perm in [ItemOLP.VIEW]:  # will have more later
-            remove_perm(item_perm, group, items_requiring_higher_trust)
-            # assign_perm accepts QuerySets for bulk assignment, but guardian just under-types it
-            # therefore, ignore the issue.
-            # See https://django-guardian.readthedocs.io/en/stable/userguide/remove/#for-multiple-objects
-            assign_perm(item_perm, group, items_requiring_lower_trust)  # type: ignore[arg-type]
+        for item in items_of_user:
+            item.recompute_group_visibility()
```
ajk8, line 185: This replaces what used to be two bulk `Item.objects.filter(...)` querysets + one bulk `remove_perm`/`assign_perm` pair (cost independent of item count) with a per-item call to `recompute_group_visibility()`, which itself loops `remove_perm` once per group currently holding VIEW on that item. For a user with many items across many groups, an ACTIVE membership save (join/leave/reactivate) is now roughly O(items × groups-with-perms) queries instead of O(1). Probably fine at current scale, but worth a benchmark or a bulk-revoke helper if group/item counts grow — this signal fires on every membership save, not just sharing changes.

borrowd_items/migrations/0023_remove_item_trust_level_required_and_add_group_sharing.py
```diff
@@ -0,0 +1,35 @@
...
+    operations = [
+        migrations.RemoveField(
+            model_name="item",
+            name="trust_level_required",
+        ),
+        migrations.AddField(
+            model_name="item",
+            name="share_with_all_groups",
+            field=models.BooleanField(
+                default=True,
```
ajk8, line 21: Worth flagging explicitly (maybe in the PR description) that this has a real visibility-widening side effect for existing data: any item that previously required `trust_level_required=HIGH` becomes visible to every ACTIVE member of every one of the owner's groups the moment this migration runs, since there's no backfill of `shared_with_groups` and `share_with_all_groups` defaults to `True`. That may well be the intended behavior now that trust levels are gone entirely, but it's a meaningful, immediate change for existing users' items in prod and seems worth calling out as a deploy consideration rather than leaving implicit in the migration default.

borrowd_items/forms.py
```diff
@@ -22,9 +26,29 @@ def validate_image_size(image: UploadedFile) -> None:
...
+    def __init__(
+        self, *args: Any, user: BorrowdUser | None = None, **kwargs: Any
+    ) -> None:
+        super().__init__(*args, **kwargs)
+        if not self.instance.pk:
+            self.initial.setdefault("share_with_all_groups", True)
+        if user is not None:
+            from borrowd_groups.models import MembershipStatus
+
+            self.fields["shared_with_groups"].queryset = user.borrowd_groups.filter(  # type: ignore[attr-defined]
```
ajk8, line 38: Nothing here (or in the template/view) stops a user from submitting `share_with_all_groups=False` with an empty `shared_with_groups` selection — the item would silently end up visible to nobody. If that's not an intended "draft"/private state, might be worth a `clean()` on the form requiring at least one group when sharing is restricted.

ajk8 (commented):
Nice, thorough removal of `TrustLevel` — I grep'd the whole tree and found no leftover references to the old trust-level templates/components/views outside of historical migrations. Left a few inline notes below; the two things I'd treat as blocking before merge are the CI failures:

- **Lint job fails**: `uvx djlint templates --reformat` still produces further diffs on `templates/includes/header.html`, `templates/items/item_form.html`, and `templates/users/profile.html` in CI, even though most of the diff in those files is already djlint-style re-indentation. Worth re-running `uvx pre-commit run --all-files` locally and committing the result.
- **E2E job fails**: the Playwright run hangs after plugin registration with no test output for ~8 minutes before being cancelled, rather than failing on an assertion. Given `item_add_new_page.py`/`item_details_page.py` were changed for the new group-sharing modal, worth confirming the new modal flow isn't the cause of the hang before merging.

Also, `borrowd.dbml` still has two `Table` fields typed as `trust_level trust_level [not null]` and a comment about "assign a trust level" (around lines 35-62), even though the `Enum trust_level { ... }` block itself was removed earlier in the same file. Since those lines fall outside this diff's hunks I can't comment inline on them, but they're now referencing a type that no longer exists in this doc.

borrowd_items/forms.py
leilabl: That is a great question for Product. I thought about that use case, it might be useful. Maybe you don't want to share the item with any groups at a given moment.

borrowd_items/migrations/0023_remove_item_trust_level_required_and_add_group_sharing.py
leilabl: This is the part of the requirements in my understanding. In this section:

Data Migration Plan:

All existing items with trust_level_required = STANDARD (1) or HIGH (2) → set share_with_all_groups = True

borrowd_groups/signals.py
leilabl: Got it. Should we create a task for this?

borrowd_items/forms.py
ajk8: Makes sense as a deliberate "not shared right now" state then — I'd just make sure the UI communicates that clearly (the button already shows "Select groups" when nothing's chosen, so a user isn't left thinking it silently reverted to some default). No form validation needed if that's intentional; happy to leave this as-is pending Product confirmation.

borrowd_items/migrations/0023_remove_item_trust_level_required_and_add_group_sharing.py
ajk8: Thanks for pointing to the migration plan — that answers it, this is intended behavior per spec rather than an oversight. Might be worth a one-line comment in the migration file (or a link to the requirements doc) so future readers don't have to dig up context for why `share_with_all_groups=True` ships with no backfill. Non-blocking either way.

borrowd_groups/signals.py
ajk8: Yeah, a follow-up ticket makes more sense than blocking this PR on it — it's not a correctness issue, just a query-count concern that'll matter more as group/item counts grow. Worth noting in the ticket: the likely fix is batching `remove_perm`/`assign_perm` across all of a user's items in one shot instead of looping per-item, similar to how the old trust-level code did it.

ajk8 (approved):
Thanks for the context in the replies — the migration behavior and the empty-sharing state are both intentional (per spec / a deliberate product choice), and the signals.py perf note is fine as a follow-up ticket rather than a blocker. Approving. One thing to double check before merging: the lint and E2E CI jobs were both red as of my last review pass — worth confirming those are green (or that the E2E failure is understood/flaky) before merge.

borrowd_items/migrations/0023_remove_item_trust_level_required_and_add_group_sharing.py
bdebut01: Discussed on Tech sync -- @ike-sudo404 going to email blast existing prod users before this logic is pushed to live

PR #529, opened 2026-07-21
Fix/527 group visibility bulk permission removal

### LINKED ISSUE #527: Group creation is O(items owned by creator x groups ) due to guardian permission recompute on every membership activation
## Summary
Creating a group triggers a synchronous permission-recompute loop (`django-guardian`) that's slow enough to exceed a 30s client timeout — and while that request is in flight, it appears to starve the shared environment's capacity to serve *other* requests too. Confirmed root driver: **484 accumulated groups** on the shared e2e test account, from a total absence of any group-cleanup mechanism.

## Severity
**Severity:** High (significant impact) - blocks any flow that creates a group on the shared e2e test account, and while the slow request is in flight, subsequent unrelated requests (e.g. loading a group detail page) can also fail to complete in time. 
Real-world severity for actual users scales with account/environment permission-table size, not just the acting user's own item count.

## Expected Behavior
Creating a group should complete quickly and not degrade responsiveness for other concurrent requests.

## Actual Behavior
`POST /groups/create/` triggers `Membership` → `ACTIVE` for the creator, which synchronously loops over every item the creator owns and does two `django-guardian` permission writes per item (`borrowd_groups/signals.py: refresh_permissions_on_membership_update` → `Item.recompute_group_visibility()`). Measured on the shared cert e2e account: creating a group takes ~40-76 seconds and exceeds a 30s client timeout entirely.

**Update after gathering real numbers (via a new e2e diagnostics test):** the acting test account owns only **21 items** but is a member/moderator of **484 groups**. 21 items alone, doing 2-3 guardian calls each, shouldn't cost 40-70s unless each individual guardian query is itself slow — which points at `django-guardian`'s permission tables being bloated by the 484-group backlog, rather than pure O(items-owned) iteration count being the dominant cost.
 **Groups have no delete/cleanup mechanism at all** (unlike items) — the only way to remove one is leaving as the sole active member (which hard-deletes it server-side), and that UI path is currently disabled (`show_leave_group_button` hardcoded to `False` in `borrowd_groups/views.py`, mid-refactor). So this backlog has been growing, unchecked, for the project's entire e2e test history.

Separately, we observed that *while* a slow create-group request is in flight, a different, unrelated request (loading the newly-created group's detail page, whose own query cost is trivial for a 1-member group) can also fail to complete within its normal timeout. This is consistent with the slow request occupying a worker for its full duration on a deployment with limited worker concurrency, starving other requests rather than each view independently being slow.

## Who Is Affected?
Any user creating a group (or triggering any `Membership` → `ACTIVE` transition) : severity scales with the size of the environment's accumulated django-guardian permission data, not just the acting user's own item count. 
Confirmed via e2e: `test_create_group`, `test_edit_group`, `test_group_detail`, `test_group_invite` all fail/are slow consistently, since each creates a throwaway group as a setup step, and `test_group_detail`/`test_group_invite` additionally fail on the *following* step (loading the group detail page) even though creation itself eventually succeeds.

## Steps to Reproduce
1. On an environment/account with a large number of groups accumulated (ideally >100), log in as a user who owns at least a handful of items.
2. Navigate to `/groups/create/`, fill in the form, and click "Create group".
3. Observe the request taking 40+ seconds; observe that a request immediately following it (e.g. loading the resulting group's detail page) may also stall.

N/A — reproduced via Playwright e2e timeouts and slowness in CI (PR #526). Real counts (21 items, 484 groups) pulled from a dedicated e2e diagnostics test (`e2e_tests/tests/test_diagnostics.py`) reading counts off the UI, attached to the Allure report, since there's no direct DB access to this environment.

## Relevant Files
- `borrowd_groups/signals.py` — `refresh_permissions_on_membership_update` (~line 178-185): loops `Item.objects.filter(owner=user)` and calls `item.recompute_group_visibility()` per item on every `ACTIVE` membership transition.
- `borrowd_items/models.py` — `recompute_group_visibility()` (~line 933): `get_groups_with_perms` + one `remove_perm` per group currently permitted + one `assign_perm` call, per item.
- `borrowd_groups/signals.py` — `set_moderator_on_group_creation`: creates the initial moderator `Membership` on group creation, triggering the chain above.
- `borrowd_groups/models.py` — `add_user()`: calls `Membership.objects.create(status=ACTIVE, ...)`.
- `borrowd_groups/views.py` — `GroupDetailView.get_context_data()`: `show_leave_group_button` hardcoded `False` (mid-refactor per #523) — the only UI path that could clean up this backlog is currently disabled.

Call chain:
```
POST /groups/create/ → GroupCreateView.form_valid()
  → BorrowdGroup post_save signal → set_moderator_on_group_creation
  → group.add_user(creator, is_moderator=True) → Membership.objects.create(status=ACTIVE)
  → Membership post_save signal → refresh_permissions_on_membership_update
  → for item in Item.objects.filter(owner=user): item.recompute_group_visibility()
      → get_groups_with_perms(item) / remove_perm(...) / assign_perm(...)  — against
        guardian permission tables sized by the whole environment's group backlog
```

## Environment
- Browser & version: Chromium (Playwright, CI)
- OS: ubuntu-latest (GitHub Actions runner)
- Python version: 3.13.14
- Anything else relevant: Reproduced against the shared cert/pre-prod environment (`E2E_BASE_URL`). Confirmed counts for the e2e test account: 21 owned items, 484 groups (via `e2e_tests/tests/test_diagnostics.py`).

## Linked Feature / Recent Change
Not confirmed to be introduced by a specific recent PR — the call chain predates #523, though #523 (per-item group sharing) touched `borrowd_items/signals.py`/`models.py` and the sharing feature generally, and also left `show_leave_group_button` hardcoded `False` (removing the only way to clean up groups) as part of an in-progress refactor toward group archive/soft-delete. Discovered while diagnosing e2e failures in PR #526. An initial hypothesis that this was a `django-guardian` autocommit/transaction conflict was checked directly against the installed `django-guardian==3.3.1` source (no `autocommit` references anywhere in the package) and ruled out.

## Possible Fix (optional)
- Batch the guardian permission calls across all of a user's items instead of one Python-level loop iteration (and DB round-trip) per item.
- Only recompute visibility for items when something relevant to sharing actually changed, rather than unconditionally on every `ACTIVE` membership transition for every group.
- If a per-item loop is unavoidable, move it off the synchronous request path (background task) so it can't starve other requests while running.
- Restore (or ship the planned archive/soft-delete replacement for) group cleanup — the 484-group backlog on the shared test account is itself a contributing factor to per-query guardian cost, and currently nothing can reduce it.

Pejjp:
## Summary
Improves the group-related permission/listing hot paths that were making the e2e pipeline slow for accounts with many groups. Fix 1 removes the per-group permission delete loop during item visibility recomputation. Fix 2 removes the groups-list/header `needs_moderator` N+1 queries by using annotated/precomputed flags.

## Linked Issue(s)
Closes #527

## Type of Change
- [x] Bug fix
- [ ] New feature
- [x] Refactor / cleanup
- [ ] Documentation

## Changes Made
- Fix 1: `Item.recompute_group_visibility()` now passes the current group queryset directly to guardian `remove_perm`, using guardian's bulk removal path instead of issuing one remove operation per group.
- Fix 2: `GroupListView` now annotates active member/moderator counts and sets `group_needs_moderator` in context, so `group_list_card.html` no longer calls `membership.group.needs_moderator` for every row.
- Fix 2: `groups_needing_moderator` context processor now uses an `Exists` query instead of iterating memberships and calling the model property on every authenticated page.
- Added regression coverage that patches `BorrowdGroup.needs_moderator` to fail, proving the list/context paths use the optimized query data instead of per-row property lookups.

## How to Test
1. `DJANGO_SETTINGS_MODULE=borrowd.config.dev.django DJANGO_SECRET_KEY=local-dev-key ./.venv/bin/python manage.py test tests.test_group_based_item_permissions`
2. `DJANGO_SETTINGS_MODULE=borrowd.config.dev.django DJANGO_SECRET_KEY=local-dev-key BORROWD_BETA_ENABLED=False ./.venv/bin/python manage.py test borrowd_groups`
3. Optional scale check: render `/groups/` with 400 memberships and confirm the query count stays around 6 SELECTs instead of the previous ~806 SELECTs.

## Local Verification (Fix 1: query-count/timing)

Query-count/timing test (local sqlite, scaled to match the real account: 400 pre-existing groups, 23 items, all `share_with_all_groups=True`, measuring the exact operation `POST /groups/create/` triggers — joining one more group):

| | queries | time (local sqlite) |
|---|---|---|
| before fix | 9,000 | 2.21s |
| after fix | 187 | 0.25s |

That's ~48x fewer queries even on fast local sqlite with zero network latency. Against the real remote Postgres on cert (where each round-trip costs real network time instead of near-zero), that 9,000-vs-187 query gap is what turns into the observed 40-94s vs. sub-second.

How this was measured, in case you want to rerun it:
- A `TestCase` builds 400 groups + 23 items for one user, then wraps a final group-join in `django.test.utils.CaptureQueriesContext` to count queries and time it.
- Ran it against the fix, then temporarily swapped in the old (pre-fix) `models.py` via `git show HEAD~1:...`, reran, then restored the fix — all local, nothing pushed or touched on cert.
- Also ran the full targeted test suites (`borrowd_groups`, `borrowd_items`, `tests.test_group_based_item_permissions`, etc. — 46+16 tests) — all pass.
- Found a regression test already covering this: `tests/test_group_based_item_permissions.py::test_item_visibility_recompute_bulk_removes_group_perms`, asserting `remove_perm` is called once with the queryset rather than once per group — verified it passes (commit `390ae60`) so this exact bug can't silently come back.
- One thing worth flagging: running the full combined suite (`borrowd_groups borrowd_items tests`) shows 42 failures/3 errors — confirmed identical with or without this fix (reverted and reran to check), so they're a pre-existing cross-app test-isolation issue, not something this change introduces. Not in scope for this fix.

## Screenshots (if UI change)
Not applicable; no visual UI change intended.

## Checklist

## Risk / Notes for Reviewer
The intended permissions behavior is unchanged. The main behavior to scrutinize is that the annotated `group_needs_moderator` flag remains equivalent to `BorrowdGroup.needs_moderator` for list/header use cases. The bulk permission removal also relies on django-guardian's queryset-aware `remove_perm` path.


vguzman812:
I approve of this PR, but it needs another person to approve it since I fixed merge conflicts.

bdebut01 (approved):

PR #539, opened 2026-07-24
cache-like bulk precomptuted item state to prevent unnecessary db round trips 

### LINKED ISSUE #531: Track remaining N+1 performance hotspots in membership permissions and inventory
## Summary
Remaining N+1-style performance hotspots may slow group membership changes and the inventory page as users accumulate items, transactions, and groups.

## Severity
**Severity:** Medium

## Expected Behavior
Group membership changes and inventory page loads should use bounded query patterns and stay responsive for users with many groups/items/transactions.

## Actual Behavior
`refresh_permissions_on_membership_update()` still performs item-by-item visibility recomputation on membership updates.

`inventory_view()`'s per-card query cost (see "Confirmed Findings" below): rendering item cards issues roughly 14 queries per item card even in the simplest case (plain `AVAILABLE` items, no transactions), scaling linearly with item count. This single root cause is responsible for three separate observed symptoms, not three separate bugs:
1. **item-deletion sweep cost** (`e2e_tests/conftest.py::_delete_e2e_items`): each deleted item redirects back to `/profile/inventory/`, paying this render cost once per item against the *current, shrinking* inventory (N, N-1, N-2, ...) — this is why item cleanup dominates the tests total time (~80%) .
2. **item-creation latency** (`test_add_item.py::test_create_item`, up to ~10s observed): `ItemCreateView.get_success_url()` redirects to `/profile/inventory/` too, so a single item creation pays this exact render cost once, against whatever the account's current item count is.
3. **General inventory page load time** for any user with a non-trivial item count.

## Who Is Affected?
Users with larger inventories, many group memberships, or many active/past transactions. This also affects the shared e2e/cert test account as its data grows over time, and (per the item-creation/deletion findings above) every e2e run that creates or deletes an item.

## Steps to Reproduce
1. Create or use a user with many groups and several owned items.
2. Trigger a membership create/update, such as creating a group or approving/joining a group.
3. Measure queries/time around `refresh_permissions_on_membership_update()` and verify it loops through owned items and calls `item.recompute_group_visibility()`.
4. Create or use a user with representative inventory activity across requests, lends, borrows, and available items.
5. Load `/profile/inventory/` and capture query counts to confirm whether card rendering performs per-item/per-transaction lookups.
6. Create one more item for a user who already owns ~20-100 items and observe `/profile/inventory/` load time on the post-create redirect.

## Confirmed Findings (inventory_view)

Measured locally with a `TestCase` + `django.test.utils.CaptureQueriesContext` hitting `/profile/inventory/` for a user owning N plain `AVAILABLE` items (no transactions, no borrowers/requesters — the cheapest possible case:

| items | total queries | transaction-table queries | photo queries |
|---|---|---|---|
| 20 | 289 | 244 | 21 |
| 50 | 709 | 604 | 51 |

~14 queries per item card, scaling linearly with item count (confirmed by doubling item count and seeing proportional growth, not worse-than-linear).

Root cause, tracing `build_item_card_context()` (`borrowd_items/card_helpers.py:303`):
1. `item.get_action_context_for(user)` (`borrowd_items/models.py:246`) calls `get_current_borrower()`, `get_requesting_user()`, and `get_current_transaction_for_user()` — each calling `Transaction.objects.get(...)` query, none cached on the instance.
2. `get_actions_for()`, called from inside `get_action_context_for`, calls `get_current_transaction_for_user()` **again** — a duplicate of the query in step 1.
3. `get_banner_info_for_item()` (`card_helpers.py:101`) calls `get_current_borrower()` and `get_requesting_user()` **again** — duplicates of step 1 — plus its own separate `item.transactions.select_related("party2").filter(...).order_by(...).first()` query, a third independent way of asking essentially the same question.
4. `item.photos.first()` (`card_helpers.py:346`) bypasses the queryset's `.prefetch_related("photos")` / `"item__photos"` entirely: `.first()` on a related manager builds a fresh ordered `QuerySet` rather than reading the prefetch cache, so it issues its own query per item regardless of the prefetch upstream (`borrowd_users/views.py` — `inventory_view()`, and `build_item_cards_for_transactions()` at `borrowd_users/views.py:294`, which is where the `item__photos` prefetch is set up but subsequently not honored by `.first()`).

Beyond `card_helpers.py`, the same three getters (`get_current_borrower`, `get_requesting_user`, `get_current_transaction_for_user`) are also called redundantly *within* `borrowd_items/models.py` itself — e.g. `is_borrowable()` (line ~687), `_get_other_user_status_text()` (line ~408), `get_actions_for()`'s own AVAILABLE-item branch (line ~437), and `process_action()` (line ~709) all call one or more of them again independently. So the duplication isn't limited to the two `card_helpers.py` call sites — it's inherent to how these getters are used throughout the per-item/per-user computation, which is why per-instance memoization (see fix below) rather than a two-call-site signature change is the right scope for this fix.

Net effect: each item card independently re-derives "who's borrowing this, who requested it, what's the current transaction" many times over through several code paths that don't share results, on top of a prefetch that silently doesn't apply at the one call site that uses it.

**Confirmed via local measurement (`GET /profile/inventory/` after creating one more item, matching real account item counts):**

| items in inventory | queries | local time (near-zero latency) | at 20ms/query (illustrative remote-DB estimate) |
|---|---|---|---|
| 24 | 345 | 0.237s | ~6.9s |
| 51 | 723 | 0.395s | ~14.5s |
| 101 | 1423 | 0.772s | ~28.5s |

At 24 items — matching the real e2e account — the illustrative remote-latency estimate lands right in the observed ~10s range for item creation, without needing to invoke any other explanation .

Not applicable; this is backend/query performance behavior.

## Relevant Files
- `borrowd_groups/signals.py` — `refresh_permissions_on_membership_update()`
- `borrowd_items/models.py` — `Item.recompute_group_visibility()`, `get_action_context_for()`, `get_actions_for()`, `get_current_borrower()`, `get_requesting_user()`, `get_current_transaction_for_user()`, `is_borrowable()`, `process_action()`
- `borrowd_users/views.py` — `inventory_view()`, `build_item_cards_for_transactions()`
- `borrowd_items/card_helpers.py` — `build_item_card_context()`, `get_banner_info_for_item()`
- `borrowd_items/views.py` — `ItemCreateView.get_success_url()` (redirects to `profile-inventory`)
- `e2e_tests/conftest.py` — `_delete_e2e_items()` (redirects to `/profile/inventory/` once per deleted item)

## Environment
- Browser & version: N/A
- OS: Local development / cert-like data shape
- Python version: 3.13+
- Anything else relevant: Django 5.2, SQLite locally; remote DB latency may amplify repeated query patterns.

## Linked Feature / Recent Change
Related to #527 and PR #529. PR #529 addresses adjacent performance work: bulk permission removal inside item visibility recomputation and the groups-list/header `needs_moderator` N+1. This issue tracks remaining work intentionally left out of that PR.

## Possible Fix (optional)
For `refresh_permissions_on_membership_update()`, consider a membership-delta update instead of full item visibility recomputation:

- ACTIVE membership: grant this group's auth group `ItemOLP.VIEW` only on this user's items that should be visible to this group.
- Non-ACTIVE membership/delete: revoke `ItemOLP.VIEW` for this group's auth group only on this user's items.
- Keep full `item.recompute_group_visibility()` for item-level changes such as `share_with_all_groups`, `shared_with_groups`, ownership transfer, or data repair.

For `inventory_view()` / card rendering, now that the N+1 is confirmed (and confirmed to also drive item-creation and item-deletion-sweep latency, not just general page load):
- Memoize `get_current_borrower()`, `get_requesting_user()`, and `get_current_transaction_for_user()` on the `Item` instance itself, invalidated on `process_action()`. Since the duplication happens across many call sites throughout `models.py` (not just the two in `card_helpers.py`), instance-level memoization is safer and more complete than threading precomputed values through method signatures at just the card-rendering call sites.
- Replace `item.photos.first()` with something that honors the existing `prefetch_related` (e.g. `next(iter(item.photos.all()), None)`) instead of building a fresh ordered queryset per item.
- Longer term, consider precomputing/batching the borrower/requester/current-transaction lookup across all items in a section with a single queryset (e.g. keyed by `item_id`) rather than per-item `.get()` calls, similar in spirit to the bulk approach used for `recompute_group_visibility()` in PR #529. Memoization removes the *duplicate* queries per item; this would remove the *per-item* queries entirely, at the cost of a larger refactor.

Pejjp:
## Summary                                                                                       
Fixes the `/profile/inventory/` page N+1 issue confirmed in #531: rendering item cards was issuing ~14 duplicate/redundant queries per item (re-querying `current borrower/requester/transaction` independently in multiple code paths, plus a `.first()` call that silently bypassed prefetching). This also directly explains why the e2e item-deletion sweep and item-creation flow were slow, since both redirect to this same page. Query cost is now a small flat overhead per section plus ~2 queries per item, regardless of how many items are in the account.                                  

## Linked Issue(s)
Closes #531                                                                                      

## Type of Change
- [ ] Bug fix
- [x] Refactor / cleanup
- [ ] New feature
- [ ] Documentation

## Changes Made
- `borrowd_items/models.py` — Added `PrecomputedItemState` (current_borrower/requesting_user/current_transaction/has_active_subscription) and a new item-scoped `Item.get_current_transaction()`, unifying three previously-different "what's this item's active transaction" queries into one verified-equivalent definition. `get_action_context_for()`/`get_actions_for()`/`_get_status_text_for_user()`/`_get_other_user_status_text()` now accept an optional `precomputed` bundle instead of each independently re-querying `get_current_borrower()`/`get_requesting_user()`/`get_current_transaction_for_user()`. Comparisons like `self.owner == user` were switched to `self.owner_id == user.id` (and similarly for `party1`/`party2`/`updated_by`) to avoid resolving FK objects just to compare identity.
- `borrowd_items/card_helpers.py` — Added batch precompute functions (`_precompute_item_states_for_items`, `_precompute_item_states_for_transactions`, `_precompute_item_state`) that derive card state for an entire section with one `Transaction` query and one `AvailabilitySubscription` query, instead of per-item lookups. Added `with_card_relations()` / `with_card_relations_for_transactions()` — queryset decorators callers apply before evaluating their queryset, so `select_related`/`prefetch_related` is set up correctly at the source . Fixed `item.photos.first()` (bypassed the applied `prefetch_related`) to `next(iter(item.photos.all()), None)`.
- `borrowd_users/views.py` (`inventory_view`) — All 5 section querysets now wrapped with `with_card_relations()`/`with_card_relations_for_transactions()`.
- `borrowd_items/views.py` (`ItemListView`, search results) — `get_queryset()` now uses `with_card_relations()` instead of a bare `.prefetch_related("photos")`.

## How to Test
1. `uv run manage.py test` — full suite (394 tests) pass unchanged; no behavior change intended.
2. Log in as a user with a non-trivial inventory (items across owned/lent/borrowed/requested), load `/profile/inventory/`, and confirm all sections, banners, and action buttons render identically to before.
3. Optional scale check: with Django Debug Toolbar or `django.test.utils.CaptureQueriesContext`, confirm inventory page query count stays flat-ish as item count grows (roughly 2 queries/item) instead of scaling at ~14/item.
4. Exercise the search page (`/items/?search=...`) to confirm item cards there are unaffected (same `build_item_cards_for_items` path, now via `with_card_relations`).

## Screenshots (if UI change)
Not applicable; no visual UI change

## Checklist
- [x] I have checked this works for r members (if relevant)

## Risk / Notes for Reviewer
Verification for this PR was done with local scripts (query capture + timing), not committed. Existing behavioral tests (`test_card_helpers.py`, borrowing/return/giveaway flow tests, etc.) all pass unchanged, which confirms the refactor doesn't affect the behaviour correctness.

The main behavior of `get_current_transaction()`'s exclusion-based status filters (`_TERMINAL_TRANSACTION_STATUSES`) is relied on to be equivalent to the old **inclusion-based** filter: verified by enumerating the full `TransactionStatus` set and excluding the non-terminal statuses:  **worth a second look**. 
Also relies on the existing app-level invariant of at most one non-terminal `Transaction` per item at a time.

### DIFF
--- a/borrowd_items/card_helpers.py
+++ b/borrowd_items/card_helpers.py
@@ -7,20 +7,25 @@ used throughout the application.
 
 from typing import TYPE_CHECKING, Any
 
+from django.db.models import QuerySet
 from django.urls import reverse
 from django.utils.html import format_html
 from django.utils.text import capfirst
 
 from .models import (
+    AvailabilitySubscription,
     AvailabilitySubscriptionStatus,
+    Item,
     ItemAction,
     ItemActionContext,
+    ItemStatus,
+    PrecomputedItemState,
+    Transaction,
+    TransactionStatus,
 )
 
 if TYPE_CHECKING:
     from borrowd_users.models import BorrowdUser
-
-    from .models import Item, Transaction
 
 
 # Giveaway listings, offers, and pending requests share the gift icon.
@@ -63,6 +68,217 @@ BANNER_STYLES = {
     "giveaway_requested": {"bg": "bg-primary/15", "text": "text-primary"},
 }
 
+_TERMINAL_TRANSACTION_STATUSES = (
+    TransactionStatus.RETURNED,
+    TransactionStatus.REJECTED,
+    TransactionStatus.CANCELLED,
+    TransactionStatus.RESOLVED,
+    TransactionStatus.OWNERSHIP_TRANSFERRED,
+)
+
+_REQUEST_TRANSACTION_STATUSES = (
+    TransactionStatus.REQUESTED,
+    TransactionStatus.GIVEAWAY_REQUESTED,
+)
+
+_BORROWER_TRANSACTION_STATUSES = (
+    TransactionStatus.ACCEPTED,
+    TransactionStatus.COLLECTION_ASSERTED,
+    TransactionStatus.COLLECTED,
+    TransactionStatus.GIVEAWAY_OFFERED,
+    TransactionStatus.RETURN_REQUESTED,
+    TransactionStatus.RETURN_ASSERTED,
+    TransactionStatus.DISPUTED,
+)
+
+_TRANSACTION_SELECT_RELATED = (
+    "party1",
+    "party1__profile",
+    "party2",
+    "party2__profile",
+    "updated_by",
+)
+
+_ITEM_SELECT_RELATED = ("owner", "owner__profile")
+
+_CARD_TRANSACTION_SELECT_RELATED = (
+    *_TRANSACTION_SELECT_RELATED,
+    "item",
+    "item__owner",
+    "item__owner__profile",
+)
+
+
+def with_card_relations(queryset: "QuerySet[Item]") -> "QuerySet[Item]":
+    """
+    Apply the select_related/prefetch_related an item card needs.
+
+    Callers building a queryset of Items destined for
+    build_item_cards_for_items() should wrap it with this before
+    evaluating it, so owner/profile/photo access doesn't cost a query
+    per card. If a caller forgets, cards still render correctly -- just
+    with a query per card instead of one for the whole list.
+    """
+    return queryset.select_related(*_ITEM_SELECT_RELATED).prefetch_related("photos")
+
+
+def with_card_relations_for_transactions(
+    queryset: "QuerySet[Transaction]",
+) -> "QuerySet[Transaction]":
+    """
+    Apply the select_related/prefetch_related an item card needs when
+    rendering from a Transaction queryset.
+
+    Callers building a queryset of Transactions destined for
+    build_item_cards_for_transactions() should wrap it with this before
+    evaluating it. If a caller forgets, cards still render correctly --
+    just with a query per card instead of one for the whole list.
+    """
+    return queryset.select_related(*_CARD_TRANSACTION_SELECT_RELATED).prefetch_related(
+        "item__photos"
+    )
+
+
+def _state_from_transaction(
+    transaction: Transaction | None,
+    *,
+    has_active_subscription: bool = False,
+) -> PrecomputedItemState:
+    """
+    Derive card state from an already-loaded transaction without querying the db.
+
+    Both action context and banner rendering need borrower/requester/current
+    transaction state, so centralizing the derivation lets them share the same
+    in-memory facts.
+    """
+    current_borrower = None
+    requesting_user = None
+
+    if transaction is not None:
+        if transaction.status in _REQUEST_TRANSACTION_STATUSES:
+            requesting_user = transaction.party2
+        elif transaction.status in _BORROWER_TRANSACTION_STATUSES:
+            current_borrower = transaction.party2
+
+    return PrecomputedItemState(
+        current_borrower=current_borrower,
+        requesting_user=requesting_user,
+        current_transaction=transaction,
+        has_active_subscription=has_active_subscription,
+    )
+
+
+def _active_subscription_item_ids(
+    item_ids: list[int],
+    user: "BorrowdUser",
+) -> set[int]:
+    """
+    Fetch active subscription flags for all card items in one query.
+
+    Without this batch lookup, unavailable item cards can each check whether
+    the viewing user has an active notify-me subscription.
+    """
+    if not item_ids:
+        return set()
+
+    return set(
+        AvailabilitySubscription.objects.filter(
+            item_id__in=item_ids,
+            user=user,
+            status=AvailabilitySubscriptionStatus.ACTIVE,
+        ).values_list("item_id", flat=True)
+    )
+
+
+def _precompute_item_states_for_items(
+    items: list["Item"],
+    user: "BorrowdUser",
+) -> dict[int, PrecomputedItemState]:
+    """
+    Build per-item card state (cache-like) for an item list with batch queries.
+
+    This replaces repeated get_current_borrower/get_requesting_user/
+    get_current_transaction/subscription calls while rendering item grids.
+    """
+    item_ids = [item.pk for item in items]
+    if not item_ids:
+        return {}
+
+    current_transactions: dict[int, Transaction] = {}
+    for transaction in (
+        Transaction.objects.filter(item_id__in=item_ids)
+        .exclude(status__in=_TERMINAL_TRANSACTION_STATUSES)
+        .select_related(*_TRANSACTION_SELECT_RELATED)
+        .order_by("item_id", "-created_at")
+    ):
+        current_transactions.setdefault(transaction.item_id, transaction)
+
+    subscription_item_ids = _active_subscription_item_ids(item_ids, user)
+
+    return {
+        item_id: _state_from_transaction(
+            current_transactions.get(item_id),
+            has_active_subscription=item_id in subscription_item_ids,
+        )
+        for item_id in item_ids
+    }
+
+
+def _precompute_item_states_for_transactions(
+    transactions: list[Transaction],
+    user: "BorrowdUser",
+) -> dict[int, PrecomputedItemState]:
+    """
+    Build per-transaction card state (cache-like) from the loaded transaction rows.
+
+    Transaction sections already know the active transaction, so this avoids
+    asking the database to rediscover the same current transaction per card.
+    """
+    item_ids = [transaction.item_id for transaction in transactions]
+    if not item_ids:
+        return {}
+
+    subscription_item_ids = _active_subscription_item_ids(item_ids, user)
+
+    return {
+        transaction.pk: _state_from_transaction(
+            transaction,
+            has_active_subscription=transaction.item_id in subscription_item_ids,
+        )
+        for transaction in transactions
+    }
+
+
+def _precompute_item_state(
+    item: "Item",
+    user: "BorrowdUser",
+) -> PrecomputedItemState:
+    """
+    Build card state for a single item when no batch state was cached or supplied.
+
+    This preserves the standalone build_item_card_context() API while keeping
+    its action-context and banner paths on one shared state object.
+    """
+    current_transaction = item.get_current_transaction()
+    needs_subscription_state = item.owner_id != user.id and (
+        current_transaction is not None or item.status != ItemStatus.AVAILABLE
+    )
+    has_active_subscription = (
+        bool(
+            AvailabilitySubscription.objects.filter(
+                item=item,
+                user=user,
+                status=AvailabilitySubscriptionStatus.ACTIVE,
+            ).exists()
+        )
+        if needs_subscription_state
+        else False
+    )
+    return _state_from_transaction(
+        current_transaction,
+        has_active_subscription=has_active_subscription,
+    )
+
 
 def build_card_ids(context: str, pk: int) -> dict[str, str]:
     """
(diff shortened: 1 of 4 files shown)


borrowd_users/views.py
```diff
@@ -258,9 +260,9 @@ def inventory_view(request: HttpRequest) -> HttpResponse:
     user = get_authenticated_user(request)
 
     # All transactions associated with the user with status == REQUESTED (awaiting approval from someone)
-    requested_transactions = Transaction.get_requested_status_transactions_for_user(
-        user
-    ).prefetch_related("item__photos")
+    requested_transactions = with_card_relations_for_transactions(
```
Pejjp, line 263: `with_card_relations_for_transactions` pre-fetches the qs necessary fields.

borrowd_items/models.py
```diff
@@ -424,7 +473,24 @@ def get_actions_for(self, user: BorrowdUser) -> tuple[ItemAction, ...]:
         """
         # This may raise Transaction.MultipleObjectsReturned.
         # Let it propagate.
-        current_tx = self.get_current_transaction_for_user(user)
+        if precomputed is not None:
```
bdebut01, line 484: Duplicated code from line [274](https://github.com/borrowd/borrowd/pull/539/changes#diff-81c219c1b9250514a1e3ae48fbf4b15462403d477a3a1f06e429bea28334150aR274) above that could get yanked into a helper function?

Pejjp: Makes a lot of sense

borrowd_items/card_helpers.py
```diff
@@ -63,6 +68,217 @@
     "giveaway_requested": {"bg": "bg-primary/15", "text": "text-primary"},
 }
 
+_TERMINAL_TRANSACTION_STATUSES = (
+    TransactionStatus.RETURNED,
+    TransactionStatus.REJECTED,
+    TransactionStatus.CANCELLED,
+    TransactionStatus.RESOLVED,
+    TransactionStatus.OWNERSHIP_TRANSFERRED,
+)
```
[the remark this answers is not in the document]
Pejjp: Done~

ajk8:
## Code Review

Automated high-effort review (multiple independent finders + adversarial verification per finding). 7 confirmed findings.

### Correctness

1. **`borrowd_items/card_helpers.py:539`** — Non-deterministic card thumbnail
   `item.photos.first()` was replaced with `next(iter(item.photos.all()), None)`, dropping the implicit pk ordering `.first()` provides on `ItemPhoto` (which has no `Meta.ordering`). On Postgres in production, an unordered `SELECT`/prefetch has no guaranteed row order, so the "first" (cover) photo shown on a card can flip to an arbitrary photo across requests or after any write touches the photos table — even though the item's photos and upload order never changed. This won't show up in dev/tests since SQLite tends to preserve insertion order for simple scans.

2. **`borrowd_items/models.py:730`** — New transaction lookup swallows a data-corruption signal
   The new `Item.get_current_transaction()` uses `.filter(...).order_by(...).first()` and silently returns the most-recent match, unlike the sibling `get_current_transaction_for_user()` (uses `.get()`, lets `Transaction.MultipleObjectsReturned` propagate). `PrecomputedItemState.current_transaction_for_user()` (models.py:127-141) likewise just returns `None` when the cached transaction isn't the queried user's party. If a data bug ever produces two non-terminal `Transaction`s on the same `Item` (a case the codebase's own comments treat as a real, if rare, possibility), the item detail page still 500s via the old method, but every list/search/inventory card — now on the new precomputed path by default — silently shows the wrong borrower/status/action buttons instead of surfacing the corruption.

3. **`borrowd_items/models.py:730`** — Dead `except` clause
   The `try`/`except Transaction.DoesNotExist` around the `.first()` call in `get_current_transaction()` is dead code — `QuerySet.first()` never raises `DoesNotExist`, it returns `None`. Misleading for future readers comparing it to the adjacent `get_current_transaction_for_user()`, which legitimately needs that except clause because it uses `.get()`.

### Efficiency

4. **`borrowd_items/models.py:298`** — Freshly-fetched state discarded, causing double queries
   `get_action_context_for()`, when `precomputed=None`, fetches `current_borrower`/`requesting_user`/`current_tx` itself in its `else` branch, but then calls `self.get_actions_for(user, precomputed=precomputed)` — forwarding the original, still-`None` `precomputed` instead of building a `PrecomputedItemState` from the values just fetched. Every caller without a precomputed state (notably `ItemDetailView.get_context_data`, i.e. the item detail page) causes `get_actions_for` to redundantly re-run the same three queries.

5. **`borrowd_items/card_helpers.py:513`** — Detail page re-derives banner state instead of reusing it
   `build_item_card_context()` only derives `precomputed` when `action_context is None` (line 513). `ItemDetailView` passes an already-built `action_context` (views.py:277-279), so `precomputed` stays `None` and is forwarded as-is to `get_banner_info_for_item`, which falls back to its own independent `_precompute_item_state()` call. Beyond the duplicated query cost, if more than one non-terminal `Transaction` ever exists for the item, the two lookups (`get_current_transaction_for_user` for action buttons vs. unscoped `get_current_transaction()` for the banner) can resolve to different rows, so the action buttons and status banner on the same card could describe two different transactions.

6. **`borrowd_users/views.py:290`** — Inventory page: one subscription query per section
   `inventory_view` calls `build_item_cards_for_transactions`/`build_item_cards_for_items` independently for each of the 5 inventory sections, so `_active_subscription_item_ids` runs once per section — a separate `AvailabilitySubscription` query per section instead of one batched query for the whole page. A much smaller residual of the exact per-item N+1 pattern this PR sets out to eliminate, left in place at per-section granularity.

### Convention

7. **`borrowd_items/card_helpers.py:179`** — Docstring describes prior code, not current behavior
   The docstring for `_precompute_item_states_for_items` reads "This replaces repeated get_current_borrower/get_requesting_user/get_current_transaction/subscription calls..." — describing what it replaces rather than its current behavior, which is exactly what AGENTS.md's evergreen-comments rule prohibits ("describe current state, not evolution... Don't reference past PRs, removed code").

---
*Generated with [Claude Code](https://claude.com/claude-code)*

vguzman812:
I'm fine with how this looks and am willing to approve it. There is one other missed part I found, though.
`/templates/components/items/modals/_giveaway_offer_modal_content.html`:
```diff
<p> You are about to permanently transfer ownership of 
     <strong>{{ item.name }}</strong> to
  -  <strong>{{ item.get_current_borrower.first_name|capfirst|default:"the borrower" }}</strong>.
  +  <strong>{{ person_name|default:"the borrower" }}</strong>.
  </p>
```

vguzman812 (approved):

Issue #536, opened 2026-07-23
Feature 2: Messages Page


iridescent-anupam:
Parent EPIC: [[EPIC]: In-App Messaging](https://github.com/borrowd/borrowd/issues/534)

### Concept

A dedicated section in the left navigation giving users a single place to view, search, and manage all their message threads — both active and archived. Full chat content lives here; the item detail page shows summaries only.

### User Story 2.1 — Messages Entry in System Pane

As a user, I want a "Messages" section in the left navigation so I can find all my conversations in one place, even if I have hundreds of them.

**Acceptance Criteria:**

- **AC 2.1.1:** A **"Messages"** entry (chat icon) is added to the left navigation for all authenticated users.
- **AC 2.1.2:** The Messages entry displays an unread badge count when there are unread threads.
- **AC 2.1.3:** `/messages/` shows a thread list with two groupings — **Active** (open pre-request chats and in-progress transactions) and **Archived** (completed, cancelled, rejected, resolved, or closed threads).
- **AC 2.1.4:** Each thread card shows: item thumbnail + name, other party's name/avatar, last message preview, date, and a status badge.
- **AC 2.1.5:** The thread list is **paginated — 25 threads per page** — with standard pagination controls (previous / next / page numbers) at the bottom of each grouping. Active and Archived threads paginate independently.

### User Story 2.2 — Full Chat Thread View

As a user, I want to open a thread and see the full conversation with item context at the top.

**Acceptance Criteria:**

- **AC 2.2.1:** Clicking a thread card opens the full chat interface.
- **AC 2.2.2:** A compact item preview card (thumbnail, name, current transaction status) is pinned at the top. It links back to the item detail page.
- **AC 2.2.3:** Messages are displayed chronologically, oldest at top, newest at bottom.
- **AC 2.2.4:** System messages (archival notices, dispute nudge, closure) are visually distinguished from user messages.
- **AC 2.2.5:** Archived threads display a clearly labelled read-only state — the message input is hidden and replaced with the archival reason.

### User Story 2.3 — Column-Level Filter Inputs

As a user with a history of transactions, I want to filter my thread list using inline inputs on each column so I can narrow down quickly without a separate filter panel.

**Acceptance Criteria:**

- **AC 2.3.1:** Each column in the thread list has an inline filter input directly beneath the column header:
    - **Person** column → text input (filters by other party's name)
    - **Item** column → text input (filters by item name)
[...]

PR #554, opened 2026-08-18
Feat/534/535 messaging chat UI

vguzman812:
## Linked Issue(s)

Closes #

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation

## Changes Made

-
-

## How to Test

1.
2.

## Checklist

### DIFF
--- a/borrowd_messaging/tests/test_views.py
+++ b/borrowd_messaging/tests/test_views.py
@@ -2,6 +2,7 @@ from django.test import override_settings
 from django.urls import reverse
 
 from borrowd_messaging.models import Message
+from borrowd_messaging.services import MessagingService
 
 from .base import MessagingTestCase
 
@@ -49,6 +50,16 @@ class ChatThreadDetailViewTests(MessagingTestCase):
         self.client.force_login(self.borrower)
 
         self.assertContains(self.client.get(self.url), "chat-end")
+
+    def test_system_notice_renders_without_a_bubble_or_avatar(self) -> None:
+        MessagingService.post_system_message(self.thread, "This item was returned.")
+        self.client.force_login(self.borrower)
+
+        response = self.client.get(self.url)
+
+        self.assertContains(response, "This item was returned.")
+        self.assertNotContains(response, "chat-bubble")
+        self.assertNotContains(response, "ui-avatars.com")
 
     def test_lender_sees_the_thread(self) -> None:
         self.client.force_login(self.lender)
--- a/borrowd_messaging/tests/test_views.py
+++ b/borrowd_messaging/tests/test_views.py
@@ -65,6 +65,14 @@ class ChatThreadDetailViewTests(MessagingTestCase):
         self.assertNotContains(response, "chat-bubble")
         self.assertNotContains(response, "ui-avatars.com")
 
+    def test_header_names_the_item(self) -> None:
+        self.client.force_login(self.borrower)
+
+        response = self.client.get(self.url)
+
+        self.assertContains(response, self.item.name)
+        self.assertNotContains(response, "no longer available")
+
     def test_lender_sees_the_thread(self) -> None:
         self.client.force_login(self.lender)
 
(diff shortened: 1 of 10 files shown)


borrowd_messaging/tests/test_views.py
```diff
@@ -0,0 +1,746 @@
...
+        self.thread = self.make_thread()
+
+    def test_archived_page_is_read_only_and_does_not_poll(self) -> None:
+        MessagingService.archive_thread(self.thread, ArchiveReason.CLOSED)
+        self.client.force_login(self.borrower)
+
+        response = self.client.get(reverse("chat-thread-detail", args=[self.thread.pk]))
+
+        self.assertContains(response, "This conversation is archived.")
+        self.assertNotContains(response, 'name="body"')
```
ajk8: This class only asserts the *negative* case (composer/close-button absent when archived). There's no remaining test that GETs a plain **active** thread and asserts the composer (`name="body"`), the poller (`hx-trigger="every 4s"`), and the close button/modal are actually rendered. A `ChatThreadActivePageTests`-style class covering this existed earlier in the PR's history but was deleted in a later cleanup commit with no replacement — right now a regression that drops the composer/poller/close-button from `chatthread_detail.html` would pass the full suite silently.

borrowd_messaging/views.py
```diff
@@ -0,0 +1,265 @@
...
+        # No new messages since last poll? send 204. htmx does not re-render/swap on a 204.
+        # https://htmx.org/docs/#requests
+        if not newer and not chat_thread.is_archived:
+            return HttpResponse(status=204)
+
+        # The transaction state may have changed after the permission lookup.
+        is_disputed = ChatThread.objects.filter(
+            pk=chat_thread.pk,
+            transaction__status=TransactionStatus.DISPUTED,
+        ).exists()
```
ajk8: This re-query runs on every poll tick that returns new messages, even when `chat_thread.transaction_id is None` (the common pre-request case), where `is_disputed` can never be true. Could short-circuit: `is_disputed = chat_thread.transaction_id is not None and ChatThread.objects.filter(...).exists()`.

Also, minor inconsistency: the stated justification is that transaction state may have changed since the permission-check lookup — but `is_archived` (read a few lines above and again below) comes from that same cached object without a similar re-query. If staleness matters for one field it arguably matters for both.

borrowd_messaging/views.py
```diff
@@ -0,0 +1,265 @@
...
+
+class ChatThreadListView(
+    MessagingEnabledMixin,
+    LoginRequiredMixin,
+    ListView[ChatThread],
+):
+    """
+    Every conversation this user is part of, newest first.
+    Deliberately bare bones.
+    TODO: update in future pr per #536
```
ajk8: `TODO: update in future pr per #536` — AGENTS.md's comment convention calls for evergreen comments that don't reference future/past PR numbers. Minor, but worth dropping or rephrasing to describe the current state instead of a planned change.

ajk8 (commented):
Ran three independent reviews of this diff (parallel, identical prompts) and cross-checked the convergent findings against the diff/worktree directly.

**Main actionable item:** the active-thread "happy path" UI (composer, poller, close button/modal) has no positive test coverage left — all three reviews independently flagged this. Left two smaller inline nits (poll-query efficiency, a stale TODO). Nothing here blocks merge in my view, but the test gap is worth closing before this ships.

borrowd_messaging/views.py
vguzman812: resolved.

borrowd_messaging/tests/test_views.py
vguzman812: resolved. Added `ChatThreadActivePageTests.test_active_prerequest_page_has_composer_poller_and_close_action`

borrowd_messaging/views.py
vguzman812: Resolved with the recommedned short circuit. I'd rather be simple than complex here for such a small feature.

templates/groups/group_list_card.html
vguzman812: just formatting because for some reason this was failing CI.

ajk8 (approved):

PR #557, opened 2026-08-21
Additional fix to address CSRF bug

ike-sudo404:
## Summary

borrowd/config/prod/django.py (and cert/django.py) never set SECURE_PROXY_SSL_HEADER. 
Platform.sh's router terminates TLS and forwards to gunicorn over plain HTTP. Without that setting, Django's request.is_secure() returns False even though the browser is on https://app.borrowd.org.

Fix: add to borrowd/config/prod/django.py (and cert/django.py, since that env sits behind the same platform.sh proxy):
SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https") 
  
That makes is_secure() correctly return True behind the proxy, so good_origin becomes https://app.borrowd.org, matching the browser's Origin header.

## Linked Issue(s)

Closes #

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Refactor / cleanup
- [ ] Documentation

## Changes Made

-
-

## How to Test

1.
2.

## Checklist

### DIFF
--- a/borrowd/config/cert/django.py
+++ b/borrowd/config/cert/django.py
@@ -10,6 +10,10 @@ from ..env import env
 
 DEBUG = False
 
+# See borrowd/config/prod/django.py for why this is required behind
+# platform.sh's TLS-terminating router.
+SECURE_PROXY_SSL_HEADER = ("HTTP_X_FORWARDED_PROTO", "https")
+
 #################################################################################
 # Platform.sh-specific configuration
 
--- a/borrowd/config/prod/django.py
+++ b/borrowd/config/prod/django.py
@@ -10,6 +10,12 @@ from ..env import env
 
 DEBUG = False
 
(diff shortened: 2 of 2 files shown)


vguzman812:
I'd like @ajk8 's input here as well for something potentially security-related, but since upsun [strips and rewrites `X-FORWARDED-PROTO`](https://fixed.docs.upsun.com/development/headers.html#headers-overwritten-or-removed-from-requests), I would like to say this implementation is OK. Django [has some warnings](https://docs.djangoproject.com/en/5.2/ref/settings/#secure-proxy-ssl-header) about setting this security setting. I am unsure this would actually fix the issue, though. 
Take a look (or have Adam do it) at the Upsun logs around the time this happened and see what that says. Search for "Forbidden" or "403".
If you can, post the results here.

ajk8 (approved):

PR #560, opened 2026-08-21
Autosaves picture when a user uploads a profile photo

### LINKED ISSUE #545: [Bug]:  autosave profile photo
## Summary
When a user adds a profile photo, they need to manually click "save" for the picture to be uploaded. Instead, the image should be automatically saved when the user adds one. 

## Severity

**Severity:**Medium

## Expected Behavior
Selecting a photo automatically uploads/saves it as the user profile image. 

## Actual Behavior
A user needs to find and click "save" or else the image is lost. 

## Who Is Affected?
All users

## Steps to Reproduce
1. Go to My profile
2. Click Edit beside photo 
3. See preview of new photo 
4. Go to another page
5. Return to My profile

## Environment
- Browser & version:
- OS:
- Python version:
- Anything else relevant:

ike-sudo404:
## Summary
Profile photos now save automatically when selected, instead of requiring a manual  "Save" click (previously the picture was silently lost if the user navigated away before clicking save).

## Linked Issue(s)

Closes #545 

## Type of Change
- [x] Bug fix

## Changes Made

- borrowd_users/forms.py: added ProfilePhotoUploadForm (extension allowlist: jpg/jpeg/png/webp, 5MB size cap) for the new upload endpoint; hardened ProfileUpdateForm.image with the same validators as a no-JS fallback.
- borrowd_users/views.py: added upload_profile_photo_view — POST-only, saves the photo immediately and deletes any prior one, returns JSON. Mirrors the existing delete_profile_photo_view pattern (kept separate from the main profile form so other in-progress fields like bio/email aren't submitted along with it).
- borrowd_users/urls.py: new profile-upload-photo route.
- templates/users/profile.html: the avatar file input now POSTs the selected file via fetch/FormData on change, shows a toast, updates the preview, and enables the Delete button - no page reload or form submit needed.

Bug found and fixed along the way (surfaced by browser-testing the change): the existing local preview used FileReader to render any selected file into the avatar , since accept="image/*" is only a picker hint a user can bypass. A non-image file tripped the 's one-shot onerror handler, permanently swapping in the placeholder icon with no way back short of a reload, even though the server correctly rejected the file. Fixed by gating the local preview on file.type.startsWith('image/'). Also fixed the new error toasts, which were rendering green (alert-success) instead of red (alert-error).

Also fixed: borrowd_users/tests/ was missing __init__.py, silently excluding its 4 test files (39 tests) from the full manage.py test run — added it, surfacing 3 pre-existing test files that were never actually running in CI.

## How to Test

1. Go to profile page. Delete photo or upload a new photo. 
2. Choose new photo for profile. 
3. Navigate to new part of website 
4. Go back to profile to verify photo was saved

## Checklist

### DIFF
--- a/borrowd_users/forms.py
+++ b/borrowd_users/forms.py
@@ -6,6 +6,9 @@ from django.contrib.auth import get_user_model
 from django.contrib.auth.forms import UserCreationForm
 from django.contrib.auth.password_validation import validate_password
 from django.core.exceptions import ValidationError
+from django.core.files.uploadedfile import UploadedFile
+from django.core.validators import FileExtensionValidator
+from django.template.defaultfilters import filesizeformat
 
 from .models import BorrowdUser, Profile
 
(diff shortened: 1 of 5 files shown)


leilabl (approved):

pyproject.toml
```diff
@@ -36,6 +36,9 @@ dev = [
     "pyyaml>=6.0.2",
 ]
 
+[tool.ruff.lint.per-file-ignores]
+"borrowd/config/*/django.py" = ["F403", "F405"]
```
leilabl: Configure ruff to suppress F403/F405 for settings files instead of inline # noqa comments — wildcard imports are intentional in Django settings inheritance.

borrowd/config/cert/django.py
```diff
@@ -5,7 +5,7 @@
 
 from borrowd.util import decode, get_platformsh_base_url
 
-from ..base import *  # noqa: F403
```
leilabl: Removed to fix lint errors. Per file ignore `[tool.ruff.lint.per-file-ignores]` added to pyproject.toml to cover all cases.

leilabl (approved):

ajk8 (approved):

